PR #24, opened 2016-10-05
cmake config exports now "include directories" of external packages for dependent projects

neumann89:
export include directories of external packages as already committed in kinetic-devel branch


croesmann:
thanks a lot, also for #25.

PR #30, opened 2016-11-03
Added suitesparse as dependency in package.xml

neumann89:
suitesparse library is not installed automatically by rosdep, when this dependency is not given in package.xml

### DIFF
--- a/package.xml
+++ b/package.xml
@@ -40,6 +40,7 @@
   <depend>pluginlib</depend>
   <depend>roscpp</depend>
   <depend>std_msgs</depend>
+  <depend>suitesparse</depend>
   <depend>tf</depend>
   <depend>tf_conversions</depend>
   <depend>visualization_msgs</depend>


croesmann:
Thank you for this pull request.

I have still some doubts regarding the _suitesparse_ dependency in general.
The library is required by g2o since _csparse_ (contained in _suitesparse_) is utilized under the hood for defining sparse linear systems. Since it is not a direct dependency of this planner, it might be better forwarded by libg2o (or the corresponding CMake find script) directly.

However, do you have any issues while trying to compile the _teb_local_planner_ without the explicit dependency on _suitesparse_ in package.xml? 
It should be installed automatically by libg2o (see https://github.com/ros-gbp/libg2o-release/blob/master/kinetic/package.xml)

neumann89:
Oh sorry, you are right! Tried again now and suitesparse is installed as a dependency of libg2o. It was probably my mistake while using rosdep.

PR #33, opened 2016-12-14
fixed wrong matrix assignment that causes a crash of the planner

mfueller:
The patched fixes a bug that causes the planner to crash since it tries to access a non-existing element of the matrix _information_.
Furthermore, it seems that the original idea is to set the non-diagonal elements of the matrix _information_inflated_ to zero.


croesmann:
Thank you for pointing this out and for providing the patch!

PR #35, opened 2016-12-16
fixed similar wrong matrix assignment that causes a crash of the planner (pr #33)

mfueller:
The same problem arises in another method in the TebOptimalPlanner class.
(see pull request #33 )


croesmann:
thanks a lot, inattentive copy-paste issue ;-)

PR #37, opened 2017-01-05
feature(publish_plan): Publish trimmed plan

reinzor:
When visualizing the global plan that is taken into account, we would
like to visualize the trimmed path instead of the full path. Since the
max lookahead distance affects the path that is taken into account by
the planner.


reinzor:
ping

croesmann:
hi @reinzor, 

sorry for the delay, I was gone for traveling. 

Thanks a lot, good point to visualize the trimmed path.
Before merging, would it be an option to use the `transformed_plan` container instead of creating a copy to `trimmed_global_plan`? Ideally, the transformed plan should mimic the relevant part of the global plan.

PR #43, opened 2017-03-24
fixed issue when start and end are the same location

wollip:
When using teb_local_planner as a path follower to follow a circular path, there is often an issue where the robot will not move. This is because on start, the robot thinks it is already at the end. By adding an extra configurable goal tolerance, we can eliminate this from happening.


croesmann:
thanks and merged @wollip.

PR #49, opened 2017-07-14
added parameter and code to update costmap footprint if it is dynamic

procopiostein:
I added a parameter and a check in the case the robot footprint is dynamnic (an arm extending for example).

Ideally I would also change the TEB footprint, but it needs a deeper change in the code, so I only implemented it for the feasibility check.


procopiostein:
I have no idea why travis has failed, the changes are minimal. Any thoughts?

croesmann:
thanks for this pull request. looks good to me :-)
The failed build does not seem to be related to your code but to the current travis config in general, so I start merging now.

PR #57, opened 2017-10-27
update to use non deprecated pluginlib macro

mikaelarguedas:
These macros, deprecated for now 8 years, will be removed in the next ROS release (ROS Melodic)

This change will allow the code to keep compiling on future ROS versions


croesmann:
thanks a lot!

PR #69, opened 2018-03-09
[test_optim_node] Normalize marker quaternions

AlexReimann:
In the recent versions RViz got more strict about normalized quaternions.
So it complains about it, this fixes it.


AlexReimann:
I actually hoped it solves a rviz crash problem I have, but doesn't ;_;

PR #86, opened 2018-07-14
Melodic devel

lwehmeier:
ROS melodic ships with a new version of g2o (with new memory management), this allows teb to compile on melodic.
Tested with melodic on bionic on armhf and x86_64


croesmann:
hey! 
thanks a lot for the pull request!
Regarding the implementation, I personally prefer the std::move syntax for unique pointers.
Since I also needed to migrate to tf2 (since the navigation stack did so in melodic now), I did these changes directly in the code to speed up pushing a melodic release (f828c75980fd43a3af160de367c9157b5b735866).
Hopefully, this is ok for you.

Thanks again, cheers,

Christoph

PR #101, opened 2018-10-19
fixed direction inversion in backwards mode when cmd_angle_instead_rottvel is true

loewt:
a small patch for using a carlike robot with ackermann_msgs
with the original code, when cmd_angle_instead_rotvel was set to true and this script was used to create Ackermann drive msgs the conversion from twist to Ackermann happened twice, inverting the sign of the steering angle when driving backwards


croesmann:
thanks a lot, I'll port this to lunar and melodic as well.

PR #105, opened 2018-11-08
new edges for carlike robots

loewt:
EdgeAcceleration: limitation of steering rate
EdgeKinematics: limitation steering angle
EdgeInflatedObstacle: squared decay for inflation radius, high cost for min obstacle distance
EdgeVelocity: penalty for omega, if leading to a smaller turning radius than possible

### DIFF
--- a/include/teb_local_planner/g2o_types/edge_acceleration.h
+++ b/include/teb_local_planner/g2o_types/edge_acceleration.h
@@ -32,7 +32,7 @@
  *  LIABILITY, OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN
  *  ANY WAY OUT OF THE USE OF THIS SOFTWARE, EVEN IF ADVISED OF THE
  *  POSSIBILITY OF SUCH DAMAGE.
- * 
+ *
  * Notes:
  * The following class is derived from a class defined by the
  * g2o-framework. g2o is licensed under the terms of the BSD License.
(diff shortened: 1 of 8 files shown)


croesmann:
Hi @loewt,

thanks a lot for this pull request. Unfortunately, I can't merge this because it contains a lot of formatting stuff that can cause conflicts with other changes. There are also some other issues in the code that need to be resolved. But before I review the code, could you tell me if you are are still working with the planner and if you have time recreate the pull request with only the relevant changes? Otherwise, I could only examine the edge changes and see what we might able to add.

loewt:
Hi @croesmann,

yes the planner is still being used, though I am currently not actively working on it. 
But I could have a look at those edges again to make the pull request conform with the formatting style.

In general the only fundamental changes are in the edges. The rest is just following those changes, since for some edges the dimension has changed.

LotfiZ:
Hi all, i know that is quite old but i have a question concerning the acceleration edge for limiting the steering rate, from the snippet code we have : 

            const double omega1 = _measurement->angular.z;
            const double omega2 = angle_diff / dt->dt();
            const double acc_rot = (omega2 - omega1) / dt->dt();

            _error[1] = penaltyBoundToInterval(acc_rot, cfg_->robot.acc_lim_theta, cfg_->optim.penalty_epsilon);

            const double steering_angle1 = _measurement->angular.z;
            const double steering_angle2 = std::atan(cfg_->robot.wheelbase / (vel2 / omega2));
            const double steering_rate = (steering_angle2 - steering_angle1) / dt->dt();`

we notice here that `const double omega1 = _measurement->angular.z;` and `const double steering_angle1 = _measurement->angular.z;`takes the same measurement that is the angular velocity but in the case of a steering angle1 we need obviously a steering angle and not an angular velocity.

@loewt  can you explain me that please ? thank you

PR #109, opened 2018-11-30
Add possibility to weight viapoint orientations

zerodamage:
For issue/enhancement of via-point following


croesmann:
Hey Christian ;-)
Thanks for creating this pull request. Can you update me on the status? Should we merge this in?

src/homotopy_class_planner.cpp
```diff
@@ -295,7 +295,7 @@ void HomotopyClassPlanner::renewAndAnalyzeOldTebs(bool delete_detours)
 
 void HomotopyClassPlanner::updateReferenceTrajectoryViaPoints(bool all_trajectories)
 {
-  if ( (!all_trajectories && !initial_plan_) || !via_points_ || via_points_->empty() || cfg_->optim.weight_viapoint <= 0)
+  if ( (!all_trajectories && !initial_plan_) || !via_points_ || via_points_->empty() || (cfg_->optim.weight_viapoint <= 0 && cfg_->optim.weight_viapoint_orientation))
```
zerodamage, line 298: This probably needs some work...

src/optimal_planner.cpp
```diff
@@ -675,8 +675,9 @@ void TebOptimalPlanner::AddEdgesViaPoints()
   
   for (ViaPointContainer::const_iterator vp_it = via_points_->begin(); vp_it != via_points_->end(); ++vp_it)
   {
+	PoseSE2 temp = *vp_it;
     
-    int index = teb_.findClosestTrajectoryPose(*vp_it, NULL, start_pose_idx);
+    int index = teb_.findClosestTrajectoryPose(temp.position(), NULL, start_pose_idx);
```
zerodamage, line 680: Don't know if this could be done in a more optimal code style

zerodamage (commented):
Some minor comments for on/off switch

zerodamage:
Hi Christoph  :)

There are some minor things as noted above, but otherwise, it should be good to go.

In the documentation, it might be worth noting that very high weights combined with sharp corners can cause oscillation or stuck behaviors.

Best, @zerodamage

andriimaistruk:
@croesmann Hi, what the status of this pr? 
@zerodamage could you provide a example of your changes actually making TEB considering poses orientation instead of just their x,y?

zerodamage:
The changes are most easily observed in:
include/teb_local_planner/g2o_types/edge_via_point.h

Adding the line:
_error[1] = g2o::normalize_theta(bandpt->theta() - _measurement->theta());

Wherein the error in orientation is added to the graph for viapoints.

That along with all the other changes to make the orientation available throughout the code as seen in the change from Vector2 to PoseSE.

For now the orientation is calculated using atan2 between adjacent points.

Buyer beware: if the orientation weight is set very high on a non-holonomic vehicle, the solver can get stuck, due to the conflicting constraints on position and orientation.

andriimaistruk:
Thanks! I see, one more question here about error value range, as in this optimization problem we try to minimize the error - push it to zero, correct? I tried to run your changes locally, I observed g2o::normalize_theta(bandpt->theta() - _measurement->theta()); outputting negative values as well, which shouldn't happen, or am I missing something here?

zerodamage:
> Thanks! I see, one more question here about error value range, as in this optimization problem we try to minimize the error - push it to zero, correct? I tried to run your changes locally, I observed g2o::normalize_theta(bandpt->theta() - _measurement->theta()); outputting negative values as well, which shouldn't happen, or am I missing something here?

Right! I've apparently missed the norm/absolute of the error in that line

andriimaistruk:
Do you plan for this pr to be merged into teb_local_planner ros2-master?
I had to make some minor changes to run it with the galactic version (I might share them if needed)

andriimaistruk:
@zerodamage

PR #112, opened 2018-12-23
many changes to edge_{velocity,acceleration,jerk}

### LINKED ISSUE #89: Oblique trajectory instead of a straight trajectory on a holonomic robot
Hi all,

Context and Issue
---
We are working with _teb_local_planner_ on a holonomic robot.

We have an issue with the planning of **forward trajectories** because the planner always gets an **oblique trajectory** (i.e. the robot rotates 45º, then it moves obliquely, and at the end, it straightens up again according to the final orientation). This [video](https://youtu.be/9UMpLxGyv4Q) shows the robot behavior and it can help you to understand better the situation.

This behavior makes sense since the vel/acc limitations are set for `x` and `y` axis (and `theta` also), so the optimizer is obtaining that the robot can move forward obliquely faster than moving straight ahead.

Question
---
How can we get trajectories no-oblique when the target is straight ahead?

---

Thanks in advance!

pingplug:
fix #89 for different types of holonomic robot

changes:
1. use normalized velocity/acceleration/jerk (remove `max_vel_x_backwards` support)
2. add option `omni_type` for different holonomic types
3. move computeError() from `include/teb_local_planner/g2o_types/` to `src/g2o_types`
4. add `exact_arc_length` support to holonomic robots
5. add jerk edges for jerk control, add `{weight_,}_jerk_lim_{x,y,theta}` options
6. tweak default `{weight_,}_{acc,jerk}_lim_{x,y,theta}` value

This PR was (partly) used in Jingdong Robotics Challenge 2018


pingplug:
I'm adding too much stuff on these hpp files, and most part of them are the same

when get acceleration, we have to calculate velocity twice (and another three times if we want jerk optimization)

there is something not right

pingplug:
so far this code is work, but not good...

pingplug:
another question: should we use normalized velocity/acceleration?

pingplug:
yet another question: why use `non holonomic constraint` but not `abs(vy)` to limit `vy`? I found it is actually calculating `hypot(vx,vy)` in `EdgeVelocity`. When `weight_kinematics_nh` is low, this is totally wrong.

gycyr (commented):
Hello, thank you for your excellent work. I would like to know if this change can be used for non-holonomic robots.

ingjae:
This hasn't been discussed for quite some time, but I'm curious as to why the jerk isn't being applied to Tebs.

PR #121, opened 2019-03-28
Get (padded) footprint from costmap when footprint_model/type parameter is missing

corot:
Any reason why TEB doesn't use costmap's footprint? Feel free to close this if there's a reason I'm missing

PS. Ignore whitespaces... my editor removed plenty, obscuring the actual changes, that I did as much minimal as I could


corot:
OK, I understand why after reading "How Obstacle Avoidance works" tutorial.

PR #124, opened 2019-04-16
publish footprint on colliding pose

naveedhd:
This (or something of this sort) helps in introspecting the cause of `infeasible trajectories`.

Please feel free to edit the right way of integrating it in the codebase.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
 .svn
 .kdev4
 *.kdev4
+.vscode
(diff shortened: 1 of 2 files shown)


naveedhd:
My particular use case was that the footprint, probably the last pose, was tipping just a bit outside of the local costmap and was causing infeasible trajectories. 

A viz of where the planner thinks the collision is happening was helpful.

corot:
:+1: But would be better if you change the id for the colliding footprint, so it doesn't get overshadowed but the visualization of current robot pose, something like on #130. Even better would be to use another color.

croesmann:
Thanks for this PR! Since #143 partly included the expected changes, I create a follow-up PR #149 to consider different colors (and namespaces). 
However, only the robot footprint model is printed and not the costmap footprint up to now (similar to the code here).

I close this PR in favor of #149.

PR #129, opened 2019-04-23
Do not ignore robot model on the association stage

corot:
This is important mostly for polygon footprint_model; with current code, almost no obstacles are considered, as with polygon footprint_model min_obstacle_dist is typically very small, possibly equal to footprint padding on costmaps.

But could also be that I misinterpreting the code... so in that case feel free to reject this!

Please ignore white spaces when comparing, as CLion doesn't like trailing white spaces


croesmann:
thanks! I think you are right, the robot model should be included properly ;-)

PR #133, opened 2019-05-20
added prune distance to pruneGlobalPlan if robot passed via_point

mgruhler:
exposes prune_distance as a ROS parameter.

Please let me know if the allocated position in the structs does not suit you.


croesmann:
thanks for the PR! Unfortunately, there are some conflicts due to some other merged changes. Could you resolve those issues (shouldn't be a big deal)? And mabye, we should rename `prune_distance` to `global_plan_prune_distance` to comply with the other global plan related parameters.

Let me know if you are busy to update PR

mgruhler:
@croesmann, thanks for the feedback.

I'll update the PR as requested, but will only be able to do so in the week after next, as I don't have a ROS machine with me atm.
So I won't feel hurt if you resolve this yourself ;-) But it is not that urgent for me, so feel free to choose what is more suitable for you.

croesmann:
ok, I did this minor change in the last commit by myself to include this in the next release that I am going to push today. Thanks!
Note, I have not included this parameter in dynamic reconfigure in order to keep the number of parameters low ;-). However, we could add this someday when the parameter is subject to be changed frequently.

mgruhler:
Cool. Thx so much! Appreciate your work here :+1:

PR #138, opened 2019-05-30
Fix sigsegv crashes

howardcochran:
These changes fix two kinds of crashes that we have encountered.
The SIGSEGV if main thread is interrupted by boost can occur when
using TebLocalPlanner within move_base_flex.


croesmann:
thanks a lot! I will merge or cherry-pick this to melodic later including changes from the other PRs.

Issue #135, opened 2019-05-23
TEB get into local minimum and break the kinematic constraint


wdxairforce:
Hi @croesmann 
i have facing the problem as the following picture show
![TEB_local_min_break_kinematic_restriction](https://user-images.githubusercontent.com/32333474/58247803-16546080-7d8d-11e9-997e-e81230e64097.png)

![TEB_1](https://user-images.githubusercontent.com/32333474/58251284-d6de4200-7d95-11e9-801b-f2ba9e031e23.png)

after read the source code and your papers of TEB, i guess there are maybe two reasons may cause this problem. 
    1) the soft constraint of kinematic that establish the approximative least-squares optimization
    2) bad initial value when there is just a "start" and "goal" make the solver get easier into local minimun
are there any other reasons would result this problem?

this problem is similar to the following issue described, 
[https://github.com/rst-tu-dortmund/teb_local_planner/issues/61](url)
do you have any method to sloved it now?

Thanks for the awesome work!
Dongxiao Wu

Issue #115, opened 2019-01-22
TEB turns around multiple times while going towards goal. 


akashjinandra:
Hello All,
We've been using TEB for while and noticed issues where it will randomly decide to turn the robot around multiple times as it gets toward the goal. In the picture below the robot is should have straight path. The global path is always straight, but I've noticed sometimes the local plan has a loop, but normally it will straight itself out before the robot gets close. 
I would say this issue happens once out of every 20 or 30 plans. 
This screenshot is from testing in gazebo, but this happens on the real vehicle as well. 
![teb_turning_around](https://user-images.githubusercontent.com/22859817/51556209-c6917700-1e36-11e9-9b25-bb36229c6966.png)

Here is another example of it happening:
![teb_werid](https://user-images.githubusercontent.com/22859817/51581694-5d395480-1e86-11e9-8bf3-d7cc773b42b1.png)

Here is my yaml file:

TebLocalPlannerROS:

 odom_topic: odom
    
 # Trajectory
  
 teb_autosize: True
 dt_ref: 1.0
 dt_hysteresis: 0.5
 global_plan_overwrite_orientation: False
 allow_init_with_backwards_motion: True
 max_global_plan_lookahead_dist: 0.0
 feasibility_check_no_poses: 3.0
 force_reinit_new_goal_dist: 1.0
 global_plan_viapoint_sep: 0.5
    
 # Robot
         
 max_vel_x: 1.0
 max_vel_x_backwards: 1.0
 max_vel_y: 0.0
 max_vel_theta: 0.5
 acc_lim_x: 1.0
 acc_lim_y: 0
 acc_lim_theta: 0.05

 # ********************** Carlike robot parameters ********************
 min_turning_radius: 0.15        # Min turning radius of the carlike robot (compute value using a model or adjust with rqt_reconfigure manually)
 wheelbase: 1.793                 # Wheelbase of our robot
 cmd_angle_instead_rotvel: False # stage simulator takes the angle instead of the rotvel as input (twist message)
 # ********************************************************************

 footprint_model:
   type: "line"
   line_start: [0.0, 0.0] # for type "line"
   line_end: [1.2, 0.0] # for type "line"

 # GoalTolerance
    
 xy_goal_tolerance: 0.05
 yaw_goal_tolerance: 0.05
 free_goal_vel: False
    
 # Obstacles
    
 min_obstacle_dist: 0.6 # This value must also include our robot's expansion, since footprint_model is set to "line".
[...]

Issue #95, opened 2018-08-23
Obstacle avoidance problem


joehrobot:
Hi,
here is my illustration:
When robot do obstacle avoidance plan, sometime teb local planner plan a weird rotation.
like following picture, blue one is robot , black circle is a weird rotation trajectory.
![teblocalplanner](https://user-images.githubusercontent.com/20719386/44504171-e3315380-a6cc-11e8-9072-0699b0e85271.png)

this my setting for teb local planner:

TebLocalPlannerROS:

 odom_topic: odom

 # Trajectory

 teb_autosize: True
 dt_ref: 0.5
 dt_hysteresis: 0.1
 global_plan_overwrite_orientation: True
 global_plan_viapoint_sep: 0.6

 allow_init_with_backwards_motion: False
 max_global_plan_lookahead_dist: 3.0
 feasibility_check_no_poses: 5

 # Robot

 max_vel_x: 0.38
 max_vel_x_backwards: 0.2
 max_vel_y: 0.0
 max_vel_theta: 0.5
 acc_lim_x: 0.2   # define the max acc
 acc_lim_theta: 0.25   #  define how fast reach max_vel_x# 1.5
 min_turning_radius: 0.0 # diff-drive robot (can turn on place!)

 #footprint_model:
   #type: "point"
 #is_footprint_dynamic: True
 # GoalTolerance

 xy_goal_tolerance: 0.07
 yaw_goal_tolerance: 0.03
 free_goal_vel: False

 # Obstacles

 min_obstacle_dist: 0.35 #  for Jackal This value must also include our robot radius, since footprint_model is set to "point".
 include_costmap_obstacles: True
 costmap_obstacles_behind_robot_dist: 1.5
 obstacle_poses_affected: 30
 costmap_converter_plugin: ""
 costmap_converter_spin_thread: True
 costmap_converter_rate: 5
 # Optimization

 no_inner_iterations: 5
 no_outer_iterations: 4
 optimization_activate: True
 optimization_verbose: False
 penalty_epsilon: 0.1
 weight_max_vel_x: 2
 weight_max_vel_theta: 1
 weight_acc_lim_x: 1
 weight_acc_lim_theta: 1
 weight_kinematics_nh: 1000
 weight_kinematics_forward_drive: 1000
 weight_kinematics_turning_radius: 1
 weight_optimaltime: 1
 weight_viapoint: 10.0

 weight_obstacle: 50
[...]

Issue #88, opened 2018-08-04
differential model robot got stuck in a local optimal trajectory


Kanelankai:
Hello!
I downloaded the kinetic-devel package and use it for local path planning with turtlebot3 in gazebo.
I found the robot still easily got stuck in a local optimal trajectory.
In teb_local_planner ROS wiki, it said that this problem had been solve by optimizing a subset of admissible trajectories of distinctive topologies in parallel. 
I don't know whether the package which I'm using can also optimizing a subset of trajectories. 
What can I do to solve this problem?
Thank  you very much.

croesmann:
Hi,
what global planner are you using? you still need a global planner to avoid to get stuck in local minima (in the static map).
However, the parallel planning is just for avoiding local minimas w.r.t. small obstacles (e.g. humans or other robots) in the local costmap.
I would recommend to complete all the [tutorials](http://wiki.ros.org/teb_local_planner/Tutorials) and to check the [teb_local_planners_tutorials](http://wiki.ros.org/teb_local_planner_tutorials) package.

_Note, this question is better placed on [answers.ros.org](http://answers.ros.org) since it does not describe any issue and it could be interesting for other users as well._

PR #139, opened 2019-05-31
Fix stuck in local minimum / excess loops in path

howardcochran:
These two commits provide a practical solution to the problem of the
planner getting stuck in a sub-optimal local minimum. Commonly, these
local minima cause the robot to perform excess loops on its way to the
goal. The solution solves the problem probabilistically by allowing the
multi-homotopy class planner to maintain a second candidate path in the
same homotopy class as the current "best" path. This allows path
exploration to "discover" new paths in a different, potentially better
local minimum, typically leading to the most optimal such alternative
being found.

This solves the following open issues:
https://github.com/rst-tu-dortmund/teb_local_planner/issues/88
https://github.com/rst-tu-dortmund/teb_local_planner/issues/95
https://github.com/rst-tu-dortmund/teb_local_planner/issues/115
https://github.com/rst-tu-dortmund/teb_local_planner/issues/135

It may also address this issue, but I am uncertain what the author means
by "teb cannot converge the path to a good solution.":
https://github.com/rst-tu-dortmund/teb_local_planner/issues/61

The second commit in this series not only improves the chances of
getting out of a local minimum but also improves the chances of finding
a better path around obstacles because it does not keep all the
candidate slots full with only the first few homotopy classes discovered
in exploration.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,6 +2,22 @@
 Changelog for package teb_local_planner
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
+Forthcoming
+-----------
+* Rename variable instead of scoping it's usage
+* test_optim_node fix circular obstacles (thanks to dtaranta)
+* Fix shadow variable warning (thanks to Victor Lopez)
+* Use SYSTEM when including external dependencies headers (thanks to Victor Lopez)
+* Limiting the control look-ahead pose to the first that execeeds the expected look-ahead time (thanks to Marco Bassa)
+* Adding the option to shift ahead the target pose used to extract the velocity command (thanks to Marco Bassa)
+* Fixed segfault in optimal_planner.cpp when clearing graph with unallocated optimizer.
+  Fixes `#158 <https://github.com/rst-tu-dortmund/teb_local_planner/issues/158>`_.
+* Added warning if parameter optimal_time is <= 0
+* Nonlinear obstacle cost from EdgeInflatedObstacle also added to EdgeObstacle.
+  See `#140 <https://github.com/rst-tu-dortmund/teb_local_planner/issues/140>`_.
+* Fixed proper initialization of parameter obstacle_cost_exponent in case it is not loaded from the parameter server
+* Contributors: Christoph Rösmann, Marco Bassa, Victor Lopez, dtaranta
+
 0.6.13 (2019-07-01)
 -------------------
 * Avoiding h signature interpolation between coincident poses (thanks to Marco Bassa)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,8 +2,8 @@
 Changelog for package teb_local_planner
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-Forthcoming
------------
+0.6.14 (2019-10-25)
+-------------------
 * Rename variable instead of scoping it's usage
 * test_optim_node fix circular obstacles (thanks to dtaranta)
 * Fix shadow variable warning (thanks to Victor Lopez)
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -2,34 +2,32 @@
 Changelog for package teb_local_planner
 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
-0.6.14 (2019-10-25)
--------------------
-* Rename variable instead of scoping it's usage
+0.8.4 (2019-12-02)
+------------------
+* Fixed TEB autoResize if last TimeDiff is small
+* Add a rotational threshold for identifying a warm start goal
+* Contributors: Rainer Kümmerle
+
+0.8.3 (2019-10-25)
+------------------
+* Limiting the control look-ahead pose to the first that execeeds the expected look-ahead time (thanks to Marco Bassa)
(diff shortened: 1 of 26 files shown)


croesmann:
Thanks a lot, looks good to me and I was able to test this successfully. 
However, I have two minor points:
- I would like to set `selection_pruning_probability` to zero for dynamic reconfiguration in order to retain the old behavior, especially in kinetic. We can think of changing this in future distribution. However, I am going to include a value of 0.3 in the teb_local_planner_tutorials configuration files.
- What do you think of renaming `pruneTebs()` to `dropTebs()` in order to avoid confusions with the meaning of pruning in `timed_elastic_band.cpp`?

howardcochran:
> Thanks a lot, looks good to me and I was able to test this successfully.

Yay, glad to hear it.

> However, I have two minor points:
> 
> * I would like to set `selection_pruning_probability` to zero for dynamic reconfiguration in order to retain the old behavior, especially in kinetic. We can think of changing this in future distribution. However, I am going to include a value of 0.3 in the teb_local_planner_tutorials configuration files.

Sure.  I set the default to 0.0 and renamed the parameter `selection_dropping_probability`

> * What do you think of renaming `pruneTebs()` to `dropTebs()` in order to avoid confusions with the meaning of pruning in `timed_elastic_band.cpp`?

I renamed `pruneTebs()` to `randomlyDropTebs()`. I also made it return immediately if `selection_dropping_probability` is zero.

And, of course, I rebased onto current kinetic-devel.
Thank you!

howardcochran:
Since my original submission, upstream changes will break this.  I will fix this for compatibility, rebase, and resubmit shortly.

howardcochran:
OK. I have rebased and force-pushed these commits to fix a compatibility issue with
43ea03209 ("Never discarding the previous best teb in renewAndAnalyzeOldTebs")

Should be ready for merging.

croesmann:
thank you @howardcochran ! I'll test this again soon!

howardcochran:
> thank you @howardcochran ! I'll test this again soon!

Thanks. Let me know if you want additional tweaks.

AntonBock:
I'm having a problem that sounds like what you have fixed @howardcochran. As you can see in the picture, my robot gets stuck trying to go through a wall, instead of going around. Is there anyway to download the fix, or should I just wait until it gets merged?

LocalMinimum

croesmann:
@AntonBock Could you try to clone this merge request and tell me if it fixes your issue?

AntonBock:
@croesmann I have cloned the merge request into the melodic devel (had to add an #include <random> in a header file). Unfortunately, it has not fixed the issue. The planner still tries to "teleport" through the wall instead of going around it. Any idea what could cause this? Should I make a new issue?

image

maxlein:
Are you sure that teb planner knows about your local obstacles?

AntonBock:
@maxlein It seems like the planner sees the obstacles. If the obstacle is smaller, ie. just a small traffic cone, the planner will easily make a path around it. I don't know if you can see it in the picture, but the planner doesn't seem to be placing the teb_markers through the wall, although the local path is planned there. This looks like the same problem posed in the ros tutorial found here: https://wiki.ros.org/teb_local_planner/Tutorials/Obstacle%20Avoidance%20and%20Robot%20Footprint%20Model, but no solution is given.

howardcochran:
> @croesmann I have cloned the merge request into the melodic devel (had to add an #include in a header file). Unfortunately, it has not fixed the issue. The planner still tries to "teleport" through the wall instead of going around it. Any idea what could cause this? Should I make a new issue?

@AntonBock, The issue that you illustrate appears very different from what my fix addresses (more on that below).
From the images you have posted, it appears that teb was initialized with an invalid plan (one that goes through the wall), and that it failed to find a valid one. For the most part, Teb depends on being initialized with a global plan that is at least topologically valid. In other words, it works best if the plan provided by the global planner goes around obstacles. Teb will then optimze the global plan from something which may be very sub-optimal and even kinesmatically infeasible into one which is kinesmatically feasible and which optimizes various tunable critera.

If given a "straight line" plan through a wall, Teb will generally push the poses forward and backward away from the wall, but there is nothing in the optimizer graph to push the plan around the wall.

HOWEVER, it may be able to find a way around the wall when you enable its Homotopy Class Planning (by setting the enable_homotopy_class_planning to true and max_number_classes parameter >= 2). This will cause it to explore random paths, which may result in it finding one that goes around that wall. But this can take quite some time and is not guaranteed to happen.

Bottom Line: Look at the configuration of your global planner and make sure it is producing a plan that goes around the wall and that this is the plan that teb is receiving.

About my change: It helps keeps Teb from getting stuck with a sub-optimal (but valid) plan  containing one or more extra loops. In your case, however, it is not finding a valid plan to begin with, so my change is not able to help.

AntonBock:
@howardcochran thank you for the detailed response! I'll look into the global planner, and see if that fixes my issue. Sorry for the misunderstanding

howardcochran:
> @howardcochran thank you for the detailed response! I'll look into the global planner, and see if that fixes my issue. Sorry for the misunderstanding

@AntonBock, you are welcome. Best of luck!

ljburtz:
has this been merged? looks like it's exactly the fix I need for the issue of the planner producing all these loops! Thanks for this awesomeness!

amakarow:
@ros-pull-request-builder retest this please

amakarow:
I just rebased this PR to melodic and merged it. Many thanks for your contribution.

Issue #28, opened 2016-10-12
Problem moving differential drive robot through narrow corridors.


jmudit19:
Hello sir,
I am trying to use teb_local_planner for passing though narrow paths efficiently. The planner works fine for broad pathways but when I am trying to pass a Robot: 
  type : "line"
   line_start: [-0.42, -0.21] 
   line_end: [0.42, -0.21]
through a door with width ~0.7m. The robot gets stuck to this:
![screenshot from 2016-10-12 21-37-57](https://cloud.githubusercontent.com/assets/22796660/19319203/704c2374-90c9-11e6-9755-0447fe4cad2a.png)
I am currently using ros-kinetic on ubuntu 15.10.
The following are the configurations used : 

![teb1](https://cloud.githubusercontent.com/assets/22796660/19319259/c2f78df2-90c9-11e6-9a70-9713eca8008c.png)
![teb2](https://cloud.githubusercontent.com/assets/22796660/19319262/c9a4250c-90c9-11e6-9137-da4d239abe42.png)

And sometimes at this (I cannot increase the inflation radius further because it does not allow the robot to move through narrow pathways ) :
![screenshot from 2016-10-12 21-35-58](https://cloud.githubusercontent.com/assets/22796660/19319365/35eecd98-90ca-11e6-8125-673a56af6c9e.png)
These are the configurations for global and local planner:
![global](https://cloud.githubusercontent.com/assets/22796660/19319429/75a33294-90ca-11e6-98e1-875869e78088.png)
![local](https://cloud.githubusercontent.com/assets/22796660/19319452/8a3b3580-90ca-11e6-897e-4934ef694c53.png)

I have also tried the same with SBPL lattice planner. Please help!

croesmann:
The error message "trajectory is not feasible" indicates that it seems like the robot-costmap-footprint (blue rectangular, defined in the costmap param file) collides with some of the lethal-obstacles cellls slightly at some point on the trajectory.

I am going to revise the planner completely in the next three weeks since I hadn't much resources during the last weeks to investigate the recently reported issues. I'll check this one as well.

nitin5:
Dear Chris,
Please also investigate why TEB do not prefer a robot to move backward in straight line as one of the option. In narrow path ways, TEB ignores this option.
Also, I have observed that TEB oscillates between local goal points as selected from the global plan.
Thanks

croesmann:
[...]

Issue #16, opened 2016-09-02
Oscillation in corridors


doisyg:
Hello,
When following a straight global path in a narrow corridor (wrt the width of the robot), i have oscillations on the trajectory that i can't explain. These oscillations are not present when navigation in wider spaces and i don't think they are due to localization noise.
Any tips or ideas to narrow the search?

doisyg:
I was able to greatly reduce these oscillations by decreasing the inflation_dist parameter (originally 0.5, changed for 0.1). Do you think there is a way to keep the desired behavior of rounding obstacle with a safe margin (the reason for the 0.5 value), without generating trajectory oscillations in a narrow space?

For reference, here are my parameters:

```
recovery_behavior_enabled: False

TebLocalPlannerROS:

 odom_topic: /encoder/odom
 map_frame: /map

 # Trajectory

 teb_autosize: True
 dt_ref: 0.3
 dt_hysteresis: 0.1
 global_plan_overwrite_orientation: True
 max_global_plan_lookahead_dist: 3.0
 feasibility_check_no_poses: 5

 # Robot

 max_vel_x: 0.42
 max_vel_x_backwards: 0.11
 max_vel_y: 0.0
 max_vel_theta: 0.6
 acc_lim_x: 0.08
 acc_lim_theta: 0.5
 min_turning_radius: 0.0
 footprint_model: # types: "point", "circular", "two_circles", "line", "polygon"
   type: "polygon"
   radius: 0.35 # for type "circular"
   line_start: [-0.3, 0.0] # for type "line"
   line_end: [0.3, 0.0] # for type "line"
   front_offset: 0.2 # for type "two_circles"
   front_radius: 0.2 # for type "two_circles"
   rear_offset: 0.2 # for type "two_circles"
   rear_radius: 0.2 # for type "two_circles"
   vertices: [[-0.27, 0.315], [0.27, 0.315], [0.27, -0.315], [-0.27, -0.315]] # for type "polygon"

 # GoalTolerance

 xy_goal_tolerance: 0.3
 yaw_goal_tolerance: 3.2
 free_goal_vel: False

 # Obstacles

 inflation_dist: 0.1   
 min_obstacle_dist: 0.1
 include_costmap_obstacles: True
 costmap_obstacles_behind_robot_dist: 1.0
 obstacle_poses_affected: 30
 costmap_converter_plugin: ""
 costmap_converter_spin_thread: True
 costmap_converter_rate: 5

[...]

PR #140, opened 2019-05-31
Add nonlinear part to obstacle cost to improve narrow gap

howardcochran:
This solves a major problem when the robot is faced with a narrow gap
that the global planner wants to traverse. The previous straight-line
obstacle cost function does not work well when there is a sufficiently
close obstacle on the left and right that the straight lines overlap.

The sum of two straight lines with opposite slope is a horizontal line,
which has zero derivative. Thus, there is no net force from obstacles
pushing poses to the center of the gap. So other slopes in the
optimization graph tend to pull these poses onto an obstacle, causing the
planned trajectory to be rejected, resetting the planner from the global
plan. We can stay in an endless loop in this case.

Reducing the obstacle weight helps a little bit in this case because the
solver becomes more stable. However, this makes it hard to avoid turning
a corner of a non-round robot into a nearby obstacle because the
gentle straight line slope of the obstacle cost is insufficient to
overcome other forces.

The new solution is to add a nonlinear cost component that will:
1. Make the minimum cost point between left & right obstacles be
in the middle.

2. Drive the cost very high as distance to obstacle aproaches zero. This
helps prevent us from turning a corner into an obstacle. Instead,
the solver will typically pull forward or backward in a slight
turn until clear enough to turn the desired direction, much like a
human driver would.

The new parameter, obstacle_cost_exponent, controls how nonlinear
the obstacle error function will be. The default is 1.0 which preserves
existing behavior.  I have found that setting the exponent to 4.0 and
increasing weight_obstacle to 100 (from 10) leads to excellent
narrow gap performance on our robot.

This solves the following open issues:
https://github.com/rst-tu-dortmund/teb_local_planner/issues/28
https://github.com/rst-tu-dortmund/teb_local_planner/issues/16

### DIFF
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -274,6 +274,10 @@ grp_optimization.add("weight_adapt_factor", double_t, 0,
   "Some special weights (currently 'weight_obstacle') are repeatedly scaled by this factor in each outer TEB iteration (weight_new: weight_old * factor); Increasing weights iteratively instead of setting a huge value a-priori leads to better numerical conditions of the underlying optimization problem.", 
   2, 1, 100) 
 
+grp_optimization.add("obstacle_cost_exponent", double_t, 0,
+	"Exponent for nonlinear obstacle cost (cost = linear_cost * obstacle_cost_exponent). Set to 1 to disable nonlinear cost (default)",
+	1, 0.01, 100)
+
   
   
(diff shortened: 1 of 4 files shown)


croesmann:
Thanks a lot for all the PRs :). I'll review them in a few days.

croesmann:
Thanks for this PR. I reviewed the code and did some tests in simulation. We can merge this now and it's good to be deactivated by default in kinetic to not break any existing configurations.

PR #146, opened 2019-06-14
Not discarding the previous teb in renewAndAnalyzeOldTebs

nirwester:
The function renewAndAnalyzeOldTebs currently discards plans pertaining to the same homotopy class without checking which one has the lowest cost.

This PR prevents this function from discarding the last best_teb_ in the case that another plan coming chronologically before in the tebs_ vector is evaluated in the same homotopy class.

To do so the last best teb is moved at the beginning of the tebs_ vector and its homotopy class identifier is added first.

I've seen there's a TODO for always selecting the teb with the lowest cost for each homotopy class; but such a change would have been more invasive (could for ex be achieved storing the tebs in a container sorted by cost).

### DIFF
--- a/src/homotopy_class_planner.cpp
+++ b/src/homotopy_class_planner.cpp
@@ -215,10 +215,10 @@ void HomotopyClassPlanner::renewAndAnalyzeOldTebs(bool delete_detours)
   equivalence_classes_.clear();
 
   // Adding the equivalence class of the latest best_teb_ first
-  bool has_best_teb = best_teb_ && std::find(tebs_.begin(), tebs_.end(), best_teb_) != tebs_.end();
+  TebOptPlannerContainer::iterator it_best_teb = best_teb_ ? std::find(tebs_.begin(), tebs_.end(), best_teb_) : tebs_.end();
+  bool has_best_teb = it_best_teb != tebs_.end();
   if (has_best_teb)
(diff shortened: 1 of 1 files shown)


src/homotopy_class_planner.cpp
```diff
@@ -214,12 +214,22 @@ void HomotopyClassPlanner::renewAndAnalyzeOldTebs(bool delete_detours)
   // clear old h-signatures (since they could be changed due to new obstacle positions.
   equivalence_classes_.clear();
 
+  // Adding the equivalence class of the latest best_teb_ first
+  bool has_best_teb = best_teb_ && std::find(tebs_.begin(), tebs_.end(), best_teb_) != tebs_.end();
+  if (has_best_teb)
+  {
+    TebOptPlannerContainer::iterator it_best_teb = std::find(tebs_.begin(), tebs_.end(), best_teb_);
```
croesmann: Should we avoid calling `std::find(tebs_.begin(), tebs_.end(), best_teb_)` twice? Indeed the expected computational burden is low since the container is usually fairly small. What do you think?

nirwester: Sure, will update the PR :)

croesmann:
thanks, merged ;-)

PR #147, opened 2019-06-15
Allow scripts to be executable and usable by rosrun after catkin_make…

UltronDestroyer:
… install and through the catkin release process

Otherwise, scripts are installed to the catkin share directory as non-executables. With this fix, it'll allow them to be run by 'rosrun' and will be installed to the catkin BIN directory which marks them as executables on install.


UltronDestroyer:
Thanks! =)

PR #150, opened 2019-06-24
New detours detection

nirwester:
As stated in the documentation, a plan is currently considered a detour if one of this cases occurs:

   1. The trajectory contains parts that increase the distance to the goal.
       This is checked by comparing the orientations theta_i with the goal heading
       If the scalar product is below the \c threshold param, the method returns true.
  2. The trajectory consist of backwards motions at the beginning of the trajectory,
       e.g. the second pose is behind the start pose w.r.t. to the same goal heading.

This approach unfortunately fails whenever the robot has to turn around a "long" obstacle (and has no other way to reach the target): the distance to the target might initially increase to later decrease, and the heading might be completely different from the start-goal heading.
The current implementation also forces the hcp planner to ignore the orientation of the initial plans provided, which again can result in unfeasible paths when detouring obstacles (see the picture below).

bad_teb_orientation

The new discard logic uses a different approach: the start orientation of the current best-teb is considered to be the reference one; this is computed as the orientation of a vector starting in the first point of the trajectory and ending in the first pose whose distance is at least "length_start_orientation_vector" (see computeStartOrientation function).
The start orientation of the other plans is computed using the same approach: these are considered detours if their heading differs from that of the best teb of at least "detours_orientation_tolerance".

The user can now enable/disable this functionality through a parameter and choose to keep the orientation of the global plan with the parameter trajectory.global_plan_overwrite_orientation.

### DIFF
--- a/include/teb_local_planner/homotopy_class_planner.h
+++ b/include/teb_local_planner/homotopy_class_planner.h
@@ -281,15 +281,6 @@ public:
    */
   void exploreEquivalenceClassesAndInitTebs(const PoseSE2& start, const PoseSE2& goal, double dist_to_obst, const geometry_msgs::Twist* start_vel);
 
-
-  /**
-   * @brief Check all available trajectories (TEBs) for detours and delete found ones.
-   * @see TimedElasticBand::detectDetoursBackwards
-   * @param threshold Threshold paramter for allowed orientation changes (below 0 -> greater than 90 deg)
-   */
-  void deleteTebDetours(double threshold=0.0);
-
-
   /**
    * @brief Add a new Teb to the internal trajectory container, if this teb constitutes a new equivalence class. Initialize it using a generic 2D reference path
    *
@@ -487,7 +478,7 @@ public:
   /**
    * @brief Internal helper function that adds a new equivalence class to the list of known classes only if it is unique.
    * @param eq_class equivalence class that should be tested
-   * @param lock if \c true, exclude the H-signature from deletion, e.g. in deleteTebDetours().
+   * @param lock if \c true, exclude the H-signature from deletion.
    * @return \c true if the h-signature was added and no duplicate was found, \c false otherwise
    */
   bool addEquivalenceClassIfNew(const EquivalenceClassPtr& eq_class, bool lock=false);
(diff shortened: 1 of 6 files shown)


croesmann:
thanks for this PR! Looks good to me and I agree that using the orientation of the current best plan is more suited than relying on the goal heading.

Just one last case to discuss:
If I got it correctly, you want to allow paths that turn around an obstacle in case no alternative exists.
However, we do not consider the following case yet (see figure):
the obstacle continuously moves to the bottom right and the trajectory extends accordingly (at least until `max_samples` is reached) without getting erased.
On the other hand,  the straight-line alternative exists, so keeping this plan is not necessary.
What do you think about including this case?

example_detour

nirwester:
I see, maybe we could handle this deleting alternative plans that have a length (or cost / traversing time) much bigger than that of the best plan? Or can you think about other strategies to delete them?

croesmann:
yes, we could compare the traversing time, e.g. `getSumOfAllTimeDiffs()`  from [here](https://github.com/rst-tu-dortmund/teb_local_planner/blob/kinetic-devel/include/teb_local_planner/timed_elastic_band.h#L613). (Hmm, maybe we should rename this method to `getTransitionTime()` or similar).

We could mark a trajectory as a detour if `cur_teb->getSumOfAllTimeDiffs() / best_teb->getSumOfAllTimeDiffs() > factor` (excluding div by 0) with `factor` around 2 (new ros parameter).

However, iterating all time differences again for all alternatives requires some additional (minor) computational resources.

We could add a method `getApproxTransitionTime()` which returns `sizeTimeDiffs()*TimeDiff(0)` as the grid is *quasi*-uniform.

What do you think?

nirwester:
Sounds reasonable, yes :)
I would assume getSumOfAllTimeDiffs to not be computationally very demanding in comparison with other parts of the code, and I'd try that in a first iteration (I'm afraid that anomalies/inaccuracies in the uniformity of the dts could make the computation less reproducible). But should it really be relevant we can for sure try your trick :) I'll PR later or next week

croesmann:
thanks a lot, I will wait for it before pushing a new release and merging into melodic :)

nirwester:
Done, I kept a bigger default limit ratio (3.0) because I found that new bands generated "on the fly", so without many optimization cycles, tend to have a relatively high traversing time even if their length is not much longer than that of the best teb. If, after playing with it, you think that a different value would fit more use-cases, we can of course change it, or switch to a length-based logic :)

croesmann:
thanks, looks good ;-)

PR #156, opened 2019-07-09
Implements both nav_core::BaseLocalPlanner and mbf_costmap_core::CostmapController abstract interfaces

corot:
The teb_local_planner plugin can still be used both in move_base and move_base_flex (MBF), but now provides an extended interface when used on MBF. Main drawback is the dependency on _mbf_costmap_core_ and _mbf_msgs_.
We provide the following exit codes:
- SUCCESS
- NOT_INITIALIZED
- INTERNAL_ERROR
- INVALID_PATH
- NO_VALID_CMD

Probably it would make sense to return also OSCILLATION, but I let this for another PR.
Also implementing cancel can make sense, when the controller takes too long to return, e.g. #144

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,8 @@ find_package(catkin REQUIRED COMPONENTS
   message_generation
   nav_core
   nav_msgs
+  mbf_costmap_core
+  mbf_msgs
   roscpp
   std_msgs
   pluginlib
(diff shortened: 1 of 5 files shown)


croesmann:
thanks a lot, @corot. I'll review the code within the next days :-)

croesmann:
looks good to me, merged! 
Thanks @corot.
The only thing is that potential *future contributors* now also need to study some basic mbf datatypes and messages. However, I prefer to support both interfaces at the moment.

PR #162, opened 2019-07-30
On footprintCost, fail only if footprint is in collision, not outside the map or on unknown space

corot:
Actually, this was already done in one of the 2 calls to footprintCost! 

Makes sense after this change on ROS nav: https://github.com/ros-planning/navigation/pull/837


croesmann:
thanks, @corot. To my mind, this doesn't affect kinetic, or am I wrong?

Issue #92, opened 2018-08-16
Oscillations around global path while using teb local planner?


kmkharche:
I am using an ackermann drive based robot with a teb local planner. However, the robot seems to oscillate around the global path, i.e. if the robot starts off to the right of the path, then goes to the left, tries to come back and again goes to the right. I am not sure what might be causing this, since the behavior with the base local planner is pretty stable compared to this. If anyone has any suggestions, please let me know. Thanks!

fangnan99:
Hi,
My team has been encountering the same problem, and we've been trying to adjust the parameters to solve this problem. Although we haven't found a proper solution to the problem, it seems that reducing `acc_lim_theta` and `max_vel_theta` could give better results. However, by doing this the car's ability of obstacle avoidance might be affected.
If you managed to overcome this problem, could you please let me know?

kmkharche:
I agree that reducing those two values helps a bit, but the robot is not able to make a lot of the sharper turns, without having to reverse a few times, which is not ideal behavior. I thought it might be because I was setting up the base link of the robot incorrectly, but even setting it at the rear wheel correctly does not change the behavior.

croesmann:
Hi,
I was not able to reproduce it yet.
Could you please try your specific parameter settings with the [teb_local_planner_tutorials](https://github.com/rst-tu-dortmund/teb_local_planner_tutorials)
package?
In particular `roslaunch teb_local_planner_tutorials robot_carlike_in_stage.launch` and its corresponding config files. 
Does the issue occur there as well? This might help us to localize the reason.
And if so, please post the configuration such that I can reproduce it.

kmkharche:
Hi,
I will try that out. But do I use the default robot footprint, or  do I need to use my robot footprint in the teb params? Because my robot footprint will be too big to maneuver through the stage world. Also, another thing I wanted to mention was that I only have a laser scan on the robot for pose estimation. I do not have motion sensors such as imu/wheel enoder for the odometry. Will that make a significant difference to the performance? And could that be responsible for the oscillations?

fangnan99:
Hi,
I think I am in a similar situation. The vehicle I am working on has a lidar for localization and a set of encoders for speed measurement. I've been tring with and without the data from the encoder, but I don't think accurate odometry information from the encoders helps a lot.

croesmann:
[...]

PR #164, opened 2019-08-09
Adding the option to shift ahead the target pose used to extract the velocity command

nirwester:
Addresses the issue https://github.com/rst-tu-dortmund/teb_local_planner/issues/92

If the reference dt used to compute the trajectory, and consequently to calculate the velocity commands sent to the base, is small enough to be comparable with the actuation / localization delays of the system, the vehicle might oscillate around the tracked global path.

This could be mitigated increasing the reference dt, but such a solution would also imply a loss of resolution in the trajectory.
Reducing the angular acceleration can also stabilize the system, reducing its "control bandwidth", but has of course a negative impact on the overall performance.

The proposed solution allows the option to shift the reference pose used for the commands generation to the future: this increases the dt used as denominator in the commands generation and at the same time doesn't sacrifice the resolution of the trajectory or the max angular acceleration of the base.

By default, the code doesn't shift the control pose, using the pose of index  1 as reference.
In our experiments, using the pose of index 3 eliminates any oscillation of the robot.
Using a big shift is not recommanded, since this would result in the controller "cutting" part of the trajectory.

### DIFF
--- a/include/teb_local_planner/homotopy_class_planner.h
+++ b/include/teb_local_planner/homotopy_class_planner.h
@@ -193,9 +193,10 @@ public:
    * @param[out] vx translational velocity [m/s]
    * @param[out] vy strafing velocity which can be nonzero for holonomic robots [m/s]
    * @param[out] omega rotational velocity [rad/s]
+   * @param[in] look_ahead_poses index of the final pose used to compute the velocity command.
    * @return \c true if command is valid, \c false otherwise
    */
-  virtual bool getVelocityCommand(double& vx, double& vy, double& omega) const;
+  virtual bool getVelocityCommand(double& vx, double& vy, double& omega, int look_ahead_poses) const;
 
   /**
    * @brief Access current best trajectory candidate (that relates to the "best" homotopy class).
--- a/include/teb_local_planner/optimal_planner.h
+++ b/include/teb_local_planner/optimal_planner.h
@@ -211,9 +211,10 @@ public:
    * @param[out] vx translational velocity [m/s]
    * @param[out] vy strafing velocity which can be nonzero for holonomic robots[m/s] 
    * @param[out] omega rotational velocity [rad/s]
+   * @param[in] look_ahead_poses index of the final pose used to compute the velocity command.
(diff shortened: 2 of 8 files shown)


croesmann:
thanks even though I think this is just a workaround rather than a solution to #92  ;-) 
Let's merge this for now...

nirwester:
Danke :)

boilerbots:
This is similar to a local change I made however I dynamically computed dt every cycle so that it tracks the actual system performance and I removed points based on their distance from the center of the robot (also removing those behind the robot). I think a distance based approach works a little better but I am trying what you have done here.

We sometimes get a hiccup in the thread since we do not use a realtime kernel so a fixed dt can cause some problems.

nirwester:
Using a fixed distance might be tricky if the range of velocities used by the platform is very wide (when driving fast, the resulting dt could become very small).
I was lately thinking that looking ahead a defined amount of time instead of a defined amount of poses would be more robust against optimizations that fail to enforce the specified d_t reference (which in my case occurs when operating in narrow passages). Let me know if you want to try that :)

boilerbots:
Yes I think that could also help. My original change was to remove points that could be very close to the center of the robot because that caused jerky movements and turns. So a fixed distance seemed OK because I was able clear out anything in the path the is close or sometimes points we had already moved past when a new plan is created while driving.

I am up for testing things, next time I will make my changes in a way that are worth pushing back to the project.

brsbilgin:
What should be the correct definition of the parameter.When I check the source code it looks "control_look_ahead_poses".  It is not visible on dynamic_reconfigure and there is no explanation on wiki page.

nirwester:
Usually the cmd_vel is generated looking at the next pose in the trajectory and computing its Dt, Dx and Dtheta.
This parameter can be read as "Index of the pose in the trajectory used to compute cmd_vel".
Increasing it has the primary effect of increasing Dt, that is used as denominator in the command generation, consequently "smoothing" the commands, making the system less sensible to actuation and localization delays.
On the other hand, since the robot won't consider the closer poses for the command generation, a big value will make it "cut" parts of the trajectory and should be avoided.

PR #167, opened 2019-08-30
Fix init with a given plan including a rotation on spot

RainerKuemmerle:
When a TEB is initialized from a plan with turn-on-spot rotations, the TEB adds a VertexTimeDiff with dt=0. Subsequently, this leads to division by zero in, for example, EdgeVelocity.
The division by zero causes the optimization to fail completely due to the NaN in the error vector.
A g2o-file would look like this:
```
VERTEX_POSE 128 10.0959 9.54506 -0.838493
VERTEX_TIMEDIFF 129 0
VERTEX_POSE 130 10.0959 9.54506 -0.981748
```
In the above example Vertex 129 would lead to the failing optimization.

Our fix is to also consider rotation on spot for computing the initial dt of the VertexTimeDiff.
Furthermore, we add an assertion in TimedElasticBand::addTimeDiff to catch this early.

### DIFF
--- a/include/teb_local_planner/timed_elastic_band.h
+++ b/include/teb_local_planner/timed_elastic_band.h
@@ -417,13 +417,14 @@ public:
    * via the argument \c dt.
    * @param plan vector of geometry_msgs::PoseStamped
    * @param max_vel_x maximum translational velocity used for determining time differences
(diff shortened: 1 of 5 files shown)


croesmann:
oh yes, indeed! thanks a lot for the fix!

PR #171, opened 2019-09-18
Warning fixes

v-lopez:


src/optimal_planner.cpp
```diff
@@ -1249,16 +1249,17 @@ bool TebOptimalPlanner::isHorizonReductionAppropriate(const std::vector<geometry
     return false;
   
   // check if distance is at least 2m long // hardcoded for now
-  double dist = 0;
-  for (int i=1; i < teb_.sizePoses(); ++i)
   {
```
croesmann, line 1255: Thanks for your suggested changes. We could also resolve shadowing by replacing `double dist = ` by `dist = ` in line 1299. Therefore, no separate separate scope would have to be added. What do you think?

v-lopez: Yes, or even better we should rename the variable. This would ensure that if in the future any part of this code was restructured or moved around the `dist` variable would not be altered in an uexpected way.

Just let me know your preference and I'll edit it.

croesmann: your right, I am fine with renaming the variable! Thanks a lot!

croesmann:
Just for the record:
I had to remove the `SYSTEM` include in `CMakeLists.txt` to fix [this error](http://build.ros.org/job/Ndev_db__teb_local_planner__debian_buster_amd64/1/console) on debian buster (`noetic-devel`, 8d00d3fbbb81e40744f0f12ea25c2a1e3666fc51).
However, if the system include is still necessary for some platforms (which I don't know at the moment), we could try this alternative fix/workaround: `-DCMAKE_NO_SYSTEM_FROM_IMPORTED=ON`.

v-lopez:
Yeah, someone else reported a similar issue on one of our repos: https://github.com/pal-robotics/play_motion/pull/47

For us the SYSTEM option enables us to have stricter build flags on our packages, while using external packages that don't adhere to these stricter flags.

Issue #154, opened 2019-06-27
ROS 2 Integration Plans


yathartha3:
The teb_local_planner for ROS1 is very useful for the robotics community, and I am curious if you have any plans for porting it to ROS2.
I am interested in helping in the porting process, if required, or try to port it myself. However, since you are the experts on this, I would like to know what your plans are :)

croesmann:
Yes, porting the planner to ROS2 is definitely on the list.
However, I wanted to wait until ROS2 and navigation2 are more stable, so I haven't checked any progress/updates in the last few months.
But I just had a look at the navigation2 repo and it seems to be in a good state.
Also in https://github.com/ros-planning/navigation2/issues/202 a certain interest in testing the teb_local_planner with the new interface is mentioned.
However, I have not yet started to review the best practices for the Navigation2 codebase and ROS2 migration. 
A very important question comes to my mind:
How do we handle maintenance and new feature development for both ROS1 and ROS2?
Just duplicating the package to teb_local_planner2 would increase the effort substantially.
Another way would be to divide the current package into several subpackages/modules so that both ROS1 and ROS2 wrappers can share the same planner backend. However, this would also imply substantial changes to teb_local_planner1. Are you aware of any best practices? Probably, the same problem applies to many other ROS packages. I'll be investigating that soon.

ghost:
> How do we handle maintenance and new feature development for both ROS1 and ROS2?

The most common solution I've seen so far is to keep the ROS 2 version on a separate branch. That way, if you make a change in ROS 1, you can just merge it to the ROS 2 version. It's still going to be extra work because you'll frequently have merge conflicts, but at least Git will help you keep track of what they are.

> navigation2 repo and it seems to be in a good state

I believe we're getting close anyway :-)

teb_local_planner seemed like a very different approach to local planning than DWA, so it's a good test of the architecture. I'd like to fix any architecture problems in navigation2 that make it harder to implement teb than it should be.

croesmann:
many thanks, @crdelsey, for providing me with the information. A ros2 branch seems to be best suited at the moment. 

Crucial dependencies are:
- costmap_converter
I'll release costmap_converter ([this PR](https://github.com/rst-tu-dortmund/costmap_converter/pull/10)) to ros2 as soon as navigation2 is in dashing (which should be happening soon according to [rosdistro](https://github.com/ros/rosdistro/pull/21764)). 
- base_local_planner
[...]

PR #174, opened 2019-09-26
Make plugin library for navigation2

vinnamkim:
Related to https://github.com/ros-planning/navigation2/issues/931, https://github.com/rst-tu-dortmund/teb_local_planner/issues/154

 - Remove teb_controller
 - Follow nav2_core::LocalPlanner interfaces to make plugin teb_local_planner::TebLocalPlannerROS
 - Update launch script and README.md to launch TurtleBot3 simulations with Navigation2 accompanied teb_local_planner

Signed-off-by: vinnamkim <vinnam.kim@gmail.com>


SteveMacenski:
Question: 
why were the messages added to a new package next to teb in dashing? I don't have an opposition to it, just curious. If the reason was technical not being able to link, I added instructions to `ros2_docs` on it because I also found it challenging. 

https://github.com/ros2/ros2_documentation/pull/339

teb_local_planner/CMakeLists.txt
```diff
@@ -70,25 +69,6 @@ set(EXTERNAL_LIBS ${Boost_LIBRARIES} ${SUITESPARSE_LIBRARIES} ${G2O_LIBRARIES})
 include_directories(
   include
   ${Boost_INCLUDE_DIRS}
-  ${costmap_converter_INCLUDE_DIRS}
```
SteveMacenski, line 73: Nice, thanks for this

teb_local_planner/launch/teb_params.yaml
```diff
@@ -51,43 +51,43 @@ bt_navigator:
     use_sim_time: True
     bt_xml_filename: "bt_navigator.xml"
 
-dwb_controller:
+bt_navigator_rclcpp_node:
```
SteveMacenski, line 54: Shouldn't think `yaml` file only contain the controller_server code?

teb_local_planner/launch/teb_view.rviz
```diff
@@ -5,7 +5,6 @@ Panels:
     Property Tree Widget:
       Expanded:
```
SteveMacenski, line 6: I'm not sure you meant to commit your rviz config?

teb_local_planner/package.xml
```diff
@@ -46,4 +45,4 @@
   <export>
     <build_type>ament_cmake</build_type>
```
SteveMacenski, line 46: Your missing the pluginlib export, see this for reference https://github.com/ros-planning/navigation2/blob/master/nav2_navfn_planner/package.xml#L37

SteveMacenski (changes requested):

teb_local_planner/package.xml
vinnamkim: Fixed.

teb_local_planner/launch/teb_view.rviz
vinnamkim: Sure, I wanted to add `/teb_poses` and `/teb_markers` for the default visualization setting.

teb_local_planner/launch/teb_params.yaml
vinnamkim: Is it possible to overwrite only for `controller_server` part?

vinnamkim:
> Question:
> why were the messages added to a new package next to teb in dashing? I don't have an opposition to it, just curious. If the reason was technical not being able to link, I added instructions to `ros2_docs` on it because I also found it challenging.
> 
> [ros2/ros2_documentation#339](https://github.com/ros2/ros2_documentation/pull/339)

This is because I just followed how other ROS2 packages did. If you want, I'll reconstruct TEB package to include the messages.

SteveMacenski:
No need to go through the effort, I was just curious if there was a specific reason.

teb_local_planner/launch/teb_params.yaml
SteveMacenski: Well in practice all the stuff in that one massive nav2_params yaml shouldn't be done like that, they should be separate yaml files for each package. But yes you can include multiple yaml files, now how it deals with conflicts is unknown to me

teb_local_planner/launch/teb_view.rviz
SteveMacenski: Ok, there was just a bunch of changes. All good

teb_local_planner/launch/teb_params.yaml
yathartha3: Probably, we should scope the local controller (dwb/teb) parameters, eg: 

```
controller_server:
  ros__paramaters:
    teb.param1 = 1
    teb.param2 = 7

yathartha3:
Just wondering, change all ```.h``` to ```.hpp```?

croesmann:
Big thanks for your efforts @vinnamkim and @SteveMacenski to bring the teb to ros2. 
I definitely prefer the new plugin approach to the previous one and the dependencies on dwb etc. removed with it.

Regarding the new msg package: this is something that can still be discussed. I already thought about splitting the packages in ros1 as it can be advantageous for usage in a multiple-machine setup as the number of dependencies (navigation stack, libg2o, costmap_converter etc.) is large. On the other hand, having multiple packages just for a local planner could also be too fine-grained (meta-package, core planner and msgs). According to the initial ros2 version of @vinnamkim, it seams that we do not need a meta package in ros2, but can still have two packages in a single git repo and a common cmake file? So up to now, I agree to proceed this way.

@yathartha3: I know that headers in ros2 core packages use `.hpp`.
However, personally, I like to use `.h` for headers even in c++ and `.hpp` for definitions of template functions/methods. But if it is best practice for all ros2 packages to follow the `.hpp` style, I would agree.

yathartha3:
How are we doing here?

yathartha3 (changes requested):
@vinnamkim I found some dwb references in the code, e.g. in, 

_planner_interface.h:_
 ```virtual bool isTrajectoryFeasible(dwb_critics::ObstacleFootprintCritic* costmap_model, blah, blah) = 0;```

_and teb_local_planner_ros.h:_
```std::shared_ptr<dwb_critics::ObstacleFootprintCritic> costmap_model_;```

There might be more elsewhere. I understand this is from the first port (when not using plugin), when we didn't have the plugin interfaces.
However, now, I believe this plugin should be free of dwb references.

croesmann:
I am going to create a new ros2 installation next week. Probably, we can push a teb ros2 release soon.

SteveMacenski (approved):

PR #178, opened 2019-10-15
Limiting the control look-ahead pose to the first that execeeds the expected look-ahead time

nirwester:
Currently the user has the option shift the pose used for the commands generation in the future in order to increase the d_t used as denominator in the commands generation.
However, if the trajectory's optimization fails to enforce the reference d_t between its poses, the resulting "control d_t" might become much bigger than the expected one (this sometimes occurs if the platform gets very close to obstacles).

This small change stops the look-ahead process at the first pose that exceeds the expected "control d_t" = look_ahead_poses * reference_dt.


croesmann:
thanks! Cherry-picked to kinetic as well ;-)

nirwester:
Danke :)

PR #180, opened 2019-10-17
Add prevent_look_ahead_poses_near_goal to fix oscillation

howardcochran:
In testing, I found that setting the new parameter
control_look_ahead_poses to 3 improved path following
accuracy very much. However, in cases where need to arrive
at a very precise goal (as by setting xy_goal_tolerance to 0.01),
it could cause the robot to overshoot the goal and then oscillate
back and forth over the goal indefinitely.

This new parameter fixes that by preventing
control_look_ahead_poses from selecting one of the last few
poses of the path.

Since this new dynamic parameter affects the behavior of an
existing parameter, it makes sense to make the existing
one dynamic also. For completeness, I made all remaining
parameters in the trajectory group be dynamic, as there is no
reason I can see not to do that.

### DIFF
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -1,5 +1,6 @@
 #!/usr/bin/env python
 
+import math
 from dynamic_reconfigure.parameter_generator_catkin import *
 #from local_planner_limits import add_generic_localplanner_params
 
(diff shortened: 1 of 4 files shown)


corot:
Hey, thanks for the PR; very valuable. I open another PR to melodic. I suppose you don't care anymore about merging this to kinetic, right?

PR #183, opened 2019-11-20
Melodic angular diff warmstart

RainerKuemmerle:
While using TEB we (by accident) flipped orientation of the goal and the former goal pose in the TEB.
This lead to loops and other local minima in the TEB optimization which Levenberg-Marquardt cannot recover.
I added a config variable to also check the angular difference between the last pose of the TEB and the new goal. In our case we avoided looping and other effects in the TEB.

### DIFF
--- a/src/homotopy_class_planner.cpp
+++ b/src/homotopy_class_planner.cpp
@@ -405,7 +405,9 @@ void HomotopyClassPlanner::updateAllTEBs(const PoseSE2* start, const PoseSE2* go
 {
   // If new goal is too far away, clear all existing trajectories to let them reinitialize later.
   // Since all Tebs are sharing the same fixed goal pose, just take the first candidate:
-  if (!tebs_.empty() && (goal->position() - tebs_.front()->teb().BackPose().position()).norm() >= cfg_->trajectory.force_reinit_new_goal_dist)
+  if (!tebs_.empty()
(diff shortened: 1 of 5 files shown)


src/homotopy_class_planner.cpp
```diff
@@ -405,7 +405,9 @@ void HomotopyClassPlanner::updateAllTEBs(const PoseSE2* start, const PoseSE2* go
 {
   // If new goal is too far away, clear all existing trajectories to let them reinitialize later.
   // Since all Tebs are sharing the same fixed goal pose, just take the first candidate:
-  if (!tebs_.empty() && (goal->position() - tebs_.front()->teb().BackPose().position()).norm() >= cfg_->trajectory.force_reinit_new_goal_dist)
+  if (!tebs_.empty()
+      && ((goal->position() - tebs_.front()->teb().BackPose().position()).norm() >= cfg_->trajectory.force_reinit_new_goal_dist
+        || fabs(g2o::normalize_theta(goal->theta() - tebs_.front()->teb().BackPose().theta())) >= cfg_->trajectory.force_reinit_new_goal_dist))
```
croesmann: You are probably comparing against the wrong parameter here, right?

croesmann (commented):
oh, I forgot submitting my review a few days ago! Just the minor change. Otherwise, it's all good! Thank you!

src/homotopy_class_planner.cpp
RainerKuemmerle: oh sorry, copy&paste mistake...

PR #185, opened 2019-11-22
Fixed TEB autoResize if last TimeDiff is small

RainerKuemmerle:
* fix the removal of a small last TimeDiff
* add a small set of tests for autoResize

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -267,6 +267,13 @@ install(DIRECTORY
 ## Testing ##
 #############
 
+if (CATKIN_ENABLE_TESTING)
+  catkin_add_gtest(test_teb_basics test/teb_basics.cpp)
+  if(TARGET test_teb_basics)
+     target_link_libraries(test_teb_basics teb_local_planner)
+  endif()
+endif()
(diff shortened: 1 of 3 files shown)


croesmann:
Many thanks Rainer!

I just have a single minor thing in mind: Usually, the unit tests require an active `roscore` whenever access to the node handle or similar is required. I have not checked this yet for the test case here. However, the official buildfarm does not start a `roscore` for plain catkin tests and hence builds might fail.
The standard way of dealing with this is `rostest`.
I did this change for the `costmap_converter` package [here](https://github.com/rst-tu-dortmund/costmap_converter/commit/443840495f79358c749d11631e8c1373484d3fd6). 
Do you mind updating this PR or should I :-)?

RainerKuemmerle:
Currently, the tests only create an instance of TimedElasticBand which if I am not mistaken does not require an active roscore.
Do you think for future extension of the tests we should migrate to rostest in this PR?

croesmann:
oh okay, good! Let's merge this for now and migrate to rostest only when we need it.

PR #187, opened 2020-01-07
Update to work with nav2 eloquent release

jplapp:
- renamed LocalPlanner to Controller
- adapted the `configure` method
- adapted `plugin_lib_names` for behaviour tree
- modified the test config params so that nav2 will choose TEB controller (naming change of `controller_plugin_types`)

I have also included a launch script to launch only the nav2 part instead of the full turtlebot, and also moved all other changes from the nav2_params.yaml test config to the teb_params.yaml test config.  These changes are not strictly necessary, let me know if I should remove them

The controller now works fine in simulation, however I noticed to problems in testing:
- commands seem to take long to compute, there are warnings that target frequency of 20Hz cannot be reached
- some, but not all, teb configuration settings are ignored. For example `footprint_model/type` is ignored while `optimization_verbose` works fine.

If you create a new eloquent-devel branch, I can re-create this PR for it.

### DIFF
--- a/teb_local_planner/include/teb_local_planner/teb_local_planner_ros.h
+++ b/teb_local_planner/include/teb_local_planner/teb_local_planner_ros.h
@@ -44,7 +44,7 @@
 #include <rclcpp/rclcpp.hpp>
 
 // Navigation2 local planner base class and utilities
-#include <nav2_core/local_planner.hpp>
+#include <nav2_core/controller.hpp>
 
 // timed-elastic-band related classes
 #include "teb_local_planner/optimal_planner.h"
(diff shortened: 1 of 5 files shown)


croesmann:
Many thanks for this pr, I just created eloquent-devel.

I am still trying to get libg2o released for ros2 before I proceed with testing and releasing teb local planner into ros2

maxlein:
> * For example `footprint_model/type` is ignored while `optimization_verbose` works fine.

In my tests I changed the footprint params so that they are in a footprint namespace if thats any help for you.  

```
-    footprint_model/type: circular
-    footprint_model/radius: 0.17
+    footprint_model:
+      type: circular
+      radius: 0.17
```

```
nh_->declare_parameter("footprint_model.type", rclcpp::ParameterValue(model_name));
nh_->declare_parameter("footprint_model.radius", rclcpp::ParameterValue(radius));
```

jplapp:
thanks @maxlein , will try this!
created #188 for eloquent-devel branch

PR #188, opened 2020-01-07
Update to work with nav2 eloquent release

jplapp:
(re-created #187)

- renamed LocalPlanner to Controller
- adapted the `configure` method
- adapted `plugin_lib_names` for behaviour tree
- modified the test config params so that nav2 will choose TEB controller (naming change of `controller_plugin_types`)

I have also included a launch script to launch only the nav2 part instead of the full turtlebot, and also moved all other changes from the nav2_params.yaml test config to the teb_params.yaml test config.  These changes are not strictly necessary, let me know if I should remove them

The controller now works fine in simulation, however I noticed to problems in testing:
- commands seem to take long to compute, there are warnings that target frequency of 20Hz cannot be reached
- some, but not all, teb configuration settings are ignored. For example `footprint_model/type` is ignored while `optimization_verbose` works fine.

### DIFF
--- a/teb_local_planner/include/teb_local_planner/teb_local_planner_ros.h
+++ b/teb_local_planner/include/teb_local_planner/teb_local_planner_ros.h
@@ -44,7 +44,7 @@
 #include <rclcpp/rclcpp.hpp>
 
 // Navigation2 local planner base class and utilities
-#include <nav2_core/local_planner.hpp>
+#include <nav2_core/controller.hpp>
 
 // timed-elastic-band related classes
 #include "teb_local_planner/optimal_planner.h"
@@ -84,7 +84,7 @@ using CostmapROSPtr = std::shared_ptr<nav2_costmap_2d::Costmap2DROS>;
   * @brief Implements the actual abstract navigation stack routines of the teb_local_planner plugin
   * @todo Escape behavior, more efficient obstacle handling
   */
-class TebLocalPlannerROS : public nav2_core::LocalPlanner
+class TebLocalPlannerROS : public nav2_core::Controller
 {
 
 public:
@@ -107,6 +107,7 @@ public:
    */
   void configure(
     const rclcpp_lifecycle::LifecycleNode::SharedPtr & node,
+    std::string name,
     const std::shared_ptr<tf2_ros::Buffer> & tf,
     const std::shared_ptr<nav2_costmap_2d::Costmap2DROS> & costmap_ros) override;
   void activate() override;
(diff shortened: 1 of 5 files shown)


SteveMacenski:
@jplapp this looks really good. The only other thing would be to namespace the parameters by the `name` given, see DWB PR as an example https://github.com/ros-planning/navigation2/pull/1468. The value of this is allowing a user to map multiple controller plugins into the controller server. That way you can have DWB, TEB, ... to choose from. But also, you can then have multiple instances of TEB in the server if you have one for more aggressive motion for certain situations. The namespacing by the `controller_plugin_id` makes sure the correct parameters are mapped to the correct plugin type. Its a 2 step process:
- Add the `name` before all parameters in the configuration file `FollowPath.param_name`, or `TEBFollowPath.param_name`, etc
- Add the `name + "." + ` in front for all parameter declarations / get / set, ei https://github.com/ros-planning/navigation2/pull/1468/files#diff-ad84955d76c45e50dc811d6a08de2b71R58

@croesmann do you mind giving this a look?

For the parameters issue @jplapp, you cannot use `/` for parameters. The namespacing is represented by `.` now. You should just rename the parameter to not be namespaced, or update how you're reading it in and representing it in yaml

SteveMacenski:
I just took a look and it looks like none of the `footprint_model` parameters have been declared, only `get`, that's your problem. Also same with `controller_frequency`. This would explain why they're not settable. 

Edit: See my branch here: https://github.com/SteveMacenski/teb_local_planner/tree/eloquent_nav2 that fixes those issues above off of your `eloquent_changes` branch. You can pull that in and retest

SteveMacenski:
I've done some more testing, the reason its not making 20hz is that it's not by default a release build. Uncomment that out and you're good to go.

croesmann:
awesome, many thanks to you both! @SteveMacenski, I merged your changes into eloquent-devel. However, [this change](https://github.com/SteveMacenski/teb_local_planner/commit/b9e1c0714a566e95852109d01d236c917d963366#diff-cd567fd420fd64c331d215a3143b7d73R126) does not compile as it does not correspond to the [function signature](https://github.com/ros-planning/navigation2/blob/master/nav2_dwb_controller/dwb_core/include/dwb_core/trajectory_critic.hpp#L94-L97). I reverted the line, however, do we need add the parent namespace to either `name` or the parent node handle?. I can also try to test this later...

SteveMacenski:
You should add the cmake flags for boost, that’s a must. 

The changing of the configure function is in a PR right now to be merged into master. That won’t get to Eloquent, but will be in Foxy. I work on master branch. The name is required to getting parameters namespaced by controller ID so you can have multiple instances of a single controller or planner with different parameters for different behaviors in context

PR #197, opened 2020-03-09
remove normalized code

XinyuKhan:
I think "Integrate with 10 steps between each pose" means dividing direction_vec by 10. not  dividing direction_vec's unit vector(normalized direction_vec) by 10.


croesmann:
thank you. I'll cherry-pick this to melodic as well.

PR #198, opened 2020-03-23
Fixed prune path [ros2]

rdeniza:
In testing we saw disappearing via points and jumping to goal directly when robot is near the goal. To fix this bug, tf_->transform() arguments should be given correctly  when transforming robot pose into the plan frame.


SteveMacenski:
#291 supersedes

PR #199, opened 2020-03-24
add visualization for circular obstacles

arii:
This small PR provides support to publish rviz markers for obstacles of the CircularObstacle type.  It follows the form of the other obstacle publishers (point, line, and polygon).


croesmann:
thank you for this PR!

PR #205, opened 2020-04-28
Fix RobotFootprintModel visualization bug

gzwsc2007:
Fixed a bug where the visualization marker array is incorrectly modified.


croesmann:
thank you!

PR #208, opened 2020-05-29
Reserve the size of the relevant obstacles vector to avoid memory allocation

jcmonteiro:
This change becomes more significant as the amount of obstacles increases. If the size is not reserved, in gcc, the capacity starts at zero, increases to one after the first `push_back`, and then doubles every time a `push_back` would cause the number of elements to exceed the current capacity. When this happens, more memory is allocated for the vector.

I didn't benchmark to measure the scenarios where this PR is more beneficial.


croesmann:
thanks!

PR #209, opened 2020-05-30
Add a saturation strategy that preserves the expected robot motion

jcmonteiro:
This other option for saturating the twist command preserves the expected robot motion. I had a good experience using it with quadrotors a while back, and I'm using now for differential drive robots with good results. It is one method of Control Allocation, namely Direct Control Allocation. I know it's unrelated, by I've published a paper a while back on the subject, I'm adding the [link here](https://ieeexplore.ieee.org/abstract/document/7524963) (minor self-promotion...).


croesmann:
the technique totally makes sense! Thanks a lot, also for the reference to your paper ;-)

PR #210, opened 2020-05-31
New type of velocity constraint

jcmonteiro:
I know this is quite a change to the planner, but I do believe it might benefit some users; besides, it is optional.

## Why
In the application I work on, the robot must maintain a minimum distance from obstacles, which increase as the robot drives faster, to ensure a given level o safety.

## Previous Attempts
We did try to tune TEB to reach this goal. As expected, it worked very well in most scenarios. However, we had problems tuning for (1) going through doors and (2) driving through corridors.

1. Doors - depending on the path taken to reach the door, the robot would try to cross it preferring one side over the other, either (1.1) making TEB give up to avoid a possible collision (we use a footprint with 1cm padding around the robot and, naturally, we are subject to sensor noise) or (1.2) making our safety system stop the robot.
2. Corridors - depending on several factors (sensor noise, temporary localization drift due to odometry errors, final goal position, ...) I could not find a suitable configuration that could consistently force the robot to drive along the middle of the corridor, without deteriorating the optimization performance in other scenarios.

## Implementation
To address the above-mentioned behaviors, I created another velocity constraint (actually, I still don't know if it is better to treat it as a constraint or as an objective). The idea is to explicitly penalize driving fast close to obstacles, which is achieved by reducing the velocity bounds as a function of the distance to an obstacle.

### Parameters
`weight_velocity_obstacle_ratio` weight for the new constraint
`obstacle_proximity_ratio_max_vel` ratio of the maximum velocities used as an upper bound when reducing the speed due to the proximity to static obstacles
`obstacle_proximity_lower_bound` distance to a static obstacle for which the velocity should be lower
`obstacle_proximity_upper_bound` distance to a static obstacle for which the velocity should be higher

### Formula
The new upper bound, as a function of the distance to an obstacle, is computed as (pseudocode):
```c++
d_max = obstacle_proximity_upper_bound;
d_min = obstacle_proximity_lower_bound;
ratio_distance = (distance - d_min) / (d_max - d_min);
ratio_distance = min( max(ratio_distance, 0), 1 );
ratio = weight_velocity_obstacle_ratio * ratio_distance;
vel_upper_bound = vel_upper_bound * ratio;
```
Then, this is used in an error function identical to the one used for `EdgeVelocity` edges.

### Example
Sample of a plan with (mostly) the default configuration.
```yaml
max_number_classes: 1
min_obstacle_dist: 1.0
```
plan before

Sample of a plan using the proposed velocity constraint.
```yaml
max_number_classes: 1
min_obstacle_dist: 1.0
weight_velocity_obstacle_ratio: 10.0
obstacle_proximity_ratio_max_vel: 0.8
obstacle_proximity_lower_bound: 0.0
obstacle_proximity_upper_bound: 1.5
```
plan after

## Relation to Other PR
PR #209 also helps avoid the door problem when it is caused by the robot drifting away from TEBs optimal plan.

### DIFF
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -189,7 +189,21 @@ grp_obstacles.add("costmap_obstacles_behind_robot_dist",   double_t,   0,
 grp_obstacles.add("obstacle_poses_affected",    int_t,    0, 
 	"The obstacle position is attached to the closest pose on the trajectory to reduce computational effort, but take a number of neighbors into account as well", 
 	30, 0, 200)
-	
+
+# Obstacle - Velocity ratio parameters
+grp_obstacles_velocity_limit = grp_obstacles.add_group("Reduce velocity near obstacles")
+
+grp_obstacles_velocity_limit.add("obstacle_proximity_ratio_max_vel", double_t, 0,
+        "Ratio of the maximum velocities used as an upper bound when reducing the speed due to the proximity to a static obstacles",
+        1, 0, 1)
+
+grp_obstacles_velocity_limit.add("obstacle_proximity_lower_bound", double_t, 0,
+        "Distance to a static obstacle for which the velocity should be lower",
+        0, 0, 10)
+
+grp_obstacles_velocity_limit.add("obstacle_proximity_upper_bound", double_t, 0,
+        "Distance to a static obstacle for which the velocity should be higher",
+        0.5, 0, 10)
 
 # Optimization
 grp_optimization = gen.add_group("Optimization", type="tab")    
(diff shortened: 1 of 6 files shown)


croesmann:
The use cases are reasonable and the results look fantastic! Thanks for this PR and also the excellent description. For sure we can get this in.

I think it will soon be time to add a new tutorial on "Adanced Tuning", with all the parameters that are now available ;-)

jcmonteiro:
You are most welcome. The work you've put in here is impressive, and I'm happy to contribute. I can try to pitch in and help with the documentation if needed.

croesmann:
thanks :) I also merged it to noetic. 

The only minor things that I observed during quick testing:
- Users may overlook the fact that they need to change `weight_velocity_obstacle_ratio` when they make adjustments to the `grp_obstacles_velocity_limit` section of rqt_reconfigure. But I don't see any better way right now to give users a hint. Maybe a one-time warning if someone changes the default parameters but keeps `weight_velocity_obstacle_ratio=0`. But It's okay for now.
- When using multiple homotopy classes with `test_optim_node` and setting `obstacle_proximity_ratio_max_vel` to a small value, I observed oscillations between the trajectory selections. However, it was only for smaller values, so I am not sure if that's relevant in practice.

jcmonteiro:
I've also experienced some oscillations, depending on the combination of weights. I've tested with very small and with very high values for `weight_velocity_obstacle_ratio`, I would only see oscillations in these cases. To be honest, I have no idea if the range from zero to one thousand is appropriate for this weight. I'll keep you informed if I see any strange behavior since I'll be using this new velocity constraint a lot.

Issue #196, opened 2020-03-05
[ROS2 Master] SegFault on printErrorWhenNotInitialized()


maxlein:
Using the current ros2-master branch I get a segfault.  
Looks like the node when using the logger for the error message is not set:  

```
Thread 1 "controller_serv" received signal SIGSEGV, Segmentation fault.
0x0000555555b50a10 in ?? ()
(gdb) bt
#0  0x0000555555b50a10 in  ()
#1  0x00007ffff49b06de in  () at /lib/x86_64-linux-gnu/libgcc_s.so.1
#2  0x00007ffff49b0ef5 in _Unwind_Resume () at /lib/x86_64-linux-gnu/libgcc_s.so.1
#3  0x00007fffba6060f3 in teb_local_planner::TebVisualization::printErrorWhenNotInitialized() const [clone .cold.991] () at /home/kin/alpen/ros2_ws/install/teb_local_planner/lib/libteb_local_planner.so
#4  0x00007ffff78d2c03 in rclcpp::callback_group::CallbackGroup::add_subscription(std::shared_ptr<rclcpp::SubscriptionBase>) () at /opt/ros/eloquent/lib/librclcpp.so
#5  0x00007ffff4fe9258 in operator new(unsigned long) () at /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0xffe8c5ebf5b53900 in  ()
#7  0x00007fffffff7880 in  ()
#8  0x00005555558f3940 in  ()
#9  0x00007fffffff6fc0 in  ()
#10 0x00007fffffff6f80 in  ()
#11 0x00007fffffff7880 in  ()
#12 0x00007fffffff8117 in  ()
#13 0x00007fffffff8270 in  ()
#14 0x00007ffff7bc9658 in rclcpp_lifecycle::LifecycleNode::LifecycleNodeInterfaceImpl::execute_callback(unsigned int, rclcpp_lifecycle::State const&) () at /opt/ros/eloquent/lib/librclcpp_lifecycle.so
#15 0x00007ffff7bc9aa1 in rclcpp_lifecycle::LifecycleNode::LifecycleNodeInterfaceImpl::change_state(unsigned char, rclcpp_lifecycle::node_interfaces::LifecycleNodeInterface::CallbackReturn&) () at /opt/ros/eloquent/lib/librclcpp_lifecycle.so
[...]

PR #212, opened 2020-06-09
Initialize Visualization after Planner to avoid segfault [ros2-master]

dkuenster:
Starting things in the current order caused a race condition when using newer ros2 versions. This fixes https://github.com/rst-tu-dortmund/teb_local_planner/issues/196. There might be a cleaner way to fix this, but this works and should at least get the idea across. Unfortunately I can only invest limited amounts of time at the moment.


croesmann:
many thanks! merged!

PR #219, opened 2020-07-24
Ported twist-to-ackermann script to ROS 2

tobiasblass:
Ported the twist-to-ackermann conversion script to ROS 2

It's probably debatable if the script should be placed under bin or lib. If you prefer another location, please just let me know and I can change it.


amakarow:
Thanks for this PR.

PR #228, opened 2020-09-17
Fixed fatal error: mbf_costmap_core/costmap_controller.h: No such fil…

Navifra-Sally:
`fatal error: mbf_costmap_core/costmap_controller.h: No such file or directory
 #include <mbf_costmap_core/costmap_controller.h>
`

I cloned the mbf_costmap_core package with git. Then I couldn't find costmap_controller.h. So I added the include directory. Then I solved!


amakarow:
Thanks.

PR #229, opened 2020-09-21
Missing usage of free_goal_vel parameter fix

BhaktiDanve:
Hello Sir,

I have been working on teb local planner package since the past few months and have found it very useful to carry on my research in the area of Motion Planning. While working on this planner, the parameter free_goal_vel was required, but I wasn’t getting the desired output when this parameter was set to true (Refer: Observations before the code fix). 

### Parameter definition

`~<name>/free_goal_vel`(bool, default: false)
Remove the goal velocity constraint such that the robot can arrive at the goal with maximum speed

### Test scenario in Rviz

A 2D-road scenario is created in Rviz. It consists of road borders and car-like actor as obstacles. Robot footprint is polygonal. 

##### free_goal_vel parameter set to false

uncorrected_free_goal_vel_false
Figure legend:- green: local plan, red: Obstacles, red arrows: teb poses

##### free_goal_vel parameter set to true

uncorrected_free_goal_vel_true
Figure legend:- green: local plan, red: Obstacles, red arrows: teb poses

As we can see in the above figures, although the parameter free_goal_vel is set to true, the poses are reaching a zero velocity near the goal. There seems to be no effect after setting this parameter to true.

After going through your code, I found some potential fixes related to the usage of this parameter:

### **Fix 1:**

**Files:**

1. src/homotopy_class_planner.cpp

- Although the parameter free_goal_vel is passed as parameter to the plan() function, it was never used in the function definition. After fixing this missing usage, I got the desired results. Please refer Observations after the code fix regarding the same.

### **Fix 2:**

**Files:**

1. include/teb_local_planner/graph_search.h
2. include/teb_local_planner/homotopy_class_planner.h
3. include/teb_local_planner/homotopy_class_planner.hpp
4. src/graph_search.cpp

- The function declarations are appended with an additional parameter  free_goal_vel (with its default value as false in the argument list) wherever necessary
- The function setVelocityGoalFree() is called at appropriate places based on the value of the free_goal_vel parameter

##### free_goal_vel parameter set to false

free_goal_vel_false
Figure legend:- green: local plan, red: Obstacles, red arrows: teb poses

##### free_goal_vel parameter set to true

free_goal_vel_true
Figure legend:- green: local plan, red: Obstacles, red arrows: teb poses

In the figures above, once the free_goal_vel parameter is set to true, it can be observed that the teb poses now reach the goal with a non-zero velocity. 

I hope my request will be concurred.

### DIFF
--- a/include/teb_local_planner/graph_search.h
+++ b/include/teb_local_planner/graph_search.h
@@ -108,7 +108,7 @@ class GraphSearchInterface
 {
 public:
 
-  virtual void createGraph(const PoseSE2& start, const PoseSE2& goal, double dist_to_obst, double obstacle_heading_threshold, const geometry_msgs::Twist* start_velocity) = 0;
+  virtual void createGraph(const PoseSE2& start, const PoseSE2& goal, double dist_to_obst, double obstacle_heading_threshold, const geometry_msgs::Twist* start_velocity, bool free_goal_vel = false) = 0;
 
   /**
    * @brief Clear any existing graph of the homotopy class search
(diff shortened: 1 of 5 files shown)


peterweissig:
Hello,

may I ask what prevents you from adding this fix ?
Can we do something to fix the "Merging is blocked" ?

Greetings
  Peter

amakarow:
As melodic is our current default branch, can you please rebase it to melodic? I am getting failed checks when I do it manually. This would be awesome. Thank you.

BhaktiDanve:
New PR [250](https://github.com/rst-tu-dortmund/teb_local_planner/pull/250) created for melodic-devel branch.

Issue #85, opened 2018-07-10
Forbid moving backwards completely?


DingFeng0103:
Hi,

I am a big fan of your work. I have tuned `weight_kinematics_forward_drive` to maximum e.g. 1000. But is still sometimes move backwards. Is there a way to completely ban this behavior?

Thanks
Feng

DingFeng0103:
Hi,
As i was checking your code, I found `penalty_epsilon` this particular parameter has no use, I mean I couldn't found a  .cpp file that utilize it. Could you tell me what this parameter is doing? and this parameter limit my `max_vel_x_backwards` so that I have to set `max_vel_x_backwards` to be bigger than `penalty_epsilon`

DingFeng0103:
Since  I found `penalty_epsilon` has no use. I set it to a very small value like 0.00001 so that I can make `max_vel_x_backwards`  really small thus backwards doesn't satisfy time optimal condition. Problem seemed solved, though I wonder if this is the right way to do it.

varandaas:
I'm also interested in this issue. I tried your solution @DingFeng0103 , but it still tried to go backward. Can you share your parameters?

DingFeng0103:
sure, I can't promise backwards move is completely forbidden, but now it hardly happens.
```

TebLocalPlannerROS:

 odom_topic: odom
 map_frame: /odom
    
 # Trajectory
  
 teb_autosize: True
 dt_ref: 0.3
 dt_hysteresis: 0.1
 global_plan_overwrite_orientation: True
 max_global_plan_lookahead_dist: 3.0
 feasibility_check_no_poses: 5
    
 # Robot
         
 max_vel_x: 0.25
 max_vel_x_backwards: 0.01
 max_vel_theta: 0.4
 acc_lim_x: 0.06
 acc_lim_theta: 0.04
 min_turning_radius: 0.0
 footprint_model: # types: "point", "circular", "two_circles", "line", "polygon"
   type: "line"
#   radius: 0.2 # for type "circular"
   line_start: [-0.1, 0.0] # for type "line"
   line_end: [0.1, 0.0] # for type "line"
#   front_offset: 0.2 # for type "two_circles"
#   front_radius: 0.2 # for type "two_circles"
#   rear_offset: 0.2 # for type "two_circles"
#   rear_radius: 0.2 # for type "two_circles"
[...]

PR #230, opened 2020-09-21
increase allowed weight_kinematics_forward_drive

doisyg:
In order to enforce forward movement only, weight_kinematics_forward_drive needs to be set higher, this PR allows it to be set to 10000.
Solves https://github.com/rst-tu-dortmund/teb_local_planner/issues/85 for my configuration


amakarow:
Thanks.

PR #232, opened 2020-09-23
Add transform_tolerance parameter

bergercookie:
Add a parameter in the teb_planner configuration as well as dynamic_reconfigure setup for specifying the transform lookup  tolerance


amakarow:
Thank you.

Issue #226, opened 2020-09-09
Build failing for teb_local_planner package on ubuntu bionic ros foxy


ghost:
```
warry@warry-desktop:~/navigation2_ws$ colcon build --symlink-install
Starting >>> nav2_common
Starting >>> nav_2d_msgs                   
Starting >>> costmap_converter_msgs                                         
Starting >>> nav2_gazebo_spawner                                                
Finished <<< nav2_common [5.64s]                                                                   
Starting >>> nav2_msgs
Starting >>> nav2_voxel_grid                                          
Finished <<< costmap_converter_msgs [7.73s]                             
Starting >>> teb_msgs
Finished <<< nav2_voxel_grid [3.20s]                                     
Finished <<< nav2_gazebo_spawner [8.67s]                                 
Finished <<< nav_2d_msgs [9.74s]                                         
Starting >>> dwb_msgs                                                                        
Finished <<< teb_msgs [6.12s]                                                                       
Finished <<< dwb_msgs [7.30s]                                                                      
Finished <<< nav2_msgs [11.9s]                          
Starting >>> nav2_util 
Finished <<< nav2_util [2.94s]                         
Starting >>> nav_2d_utils
Starting >>> nav2_behavior_tree              
Starting >>> nav2_lifecycle_manager                                                  
Starting >>> nav2_map_server                                                             
Starting >>> nav2_amcl                                                                   
Starting >>> nav2_waypoint_follower                                                      
Finished <<< nav_2d_utils [5.31s]                                                              
Finished <<< nav2_lifecycle_manager [6.42s]                                   
Starting >>> nav2_rviz_plugins
Finished <<< nav2_waypoint_follower [6.36s]                                    
Finished <<< nav2_amcl [7.34s]                                                 
Finished <<< nav2_map_server [7.59s]                                           
Starting >>> nav2_costmap_2d
Finished <<< nav2_rviz_plugins [4.69s]                                           
Finished <<< nav2_behavior_tree [13.2s]                                           
Starting >>> nav2_bt_navigator
Finished <<< nav2_costmap_2d [6.83s]                                                                
Starting >>> nav2_core
Starting >>> costmap_queue                                                                   
Starting >>> costmap_converter                                                                   
Finished <<< nav2_core [1.97s]                                                                      
Starting >>> dwb_core                                                           
Starting >>> nav2_controller                                                    
Starting >>> nav2_navfn_planner                                                 
[...]

Issue #215, opened 2020-07-08
[ROS2] API change incoming


SteveMacenski:
Hi,

From https://github.com/ros-planning/navigation2/pull/1857 we're moving the progress and goal checkers internally to the nav2_controller server rather than having to be supplied by the algorithm. As such, we're removing the `isGoalReached` from the controller API that TEB implements. If you need a specialized goal checker plugin, please let us know and we can work to get that included into the base navigation2 stack. If you don't need a specialized one, please look over the default one and make sure that it meets your needs (was used in DWB).

croesmann:
Many thanks for letting us know! The current goal checking is very similar to the one from DWB except the changes merged in #43. The check there only addresses TEB with via-points and circular global plans. 

The use case is actually when start and goal are quite similar (w.r.t. position and orientation tolerances),
but the global planner creates a trajectory of considerable lengthnto pass some other waypoints, the current DWB goal checker still returns "goal reached".
Perhaps a more abstract way to integrate this into the goal checker in the base navigation stack would be to include the remaining length of the global plan as another goal condition. This would enable this type of application for all planners.
(The number of via points is derived from the global plan based on a predefined grid, so that the verification of the number of via points or global plan lengths is almost equivalent).
But to be honest, this a use case that does not occur often in applications ;-)

SteveMacenski:
Well the number of points can't be relied upon. The sparsity of plans can't be assumed. Some might be very dense, others might not be.

Michael-Equi:
Has a decision been reached on this? The change to the nav2 stack prevents the TEB planner form building: `error: ‘bool teb_local_planner::TebLocalPlannerROS::isGoalReached(const PoseStamped&, const Twist&)’ marked ‘override’, but does not override`.

SteveMacenski:
Another API change https://github.com/ros-planning/navigation2/blob/45775b360a8bad5e1fb04cf7fb230a318159df4b/nav2_core/include/nav2_core/controller.hpp we're moving to WeakPtrs over sharedPtrs to deal with cleaner exits in the active state.

eliasdc:
> Has a decision been reached on this? The change to the nav2 stack prevents the TEB planner form building: `error: ‘bool teb_local_planner::TebLocalPlannerROS::isGoalReached(const PoseStamped&, const Twist&)’ marked ‘override’, but does not override`.

is there a workaround to fix this?

SteveMacenski:
Submit a PR for the master branch updates

PR #233, opened 2020-10-09
Ros2 api update

SteveMacenski:
https://github.com/rst-tu-dortmund/teb_local_planner/issues/226

https://github.com/rst-tu-dortmund/teb_local_planner/issues/215

Resolves these 2 tickets:
- Removes goal checker condition, now the controller server takes care of this for you
- Changes to the new weak pointer API

I just checked that things build, but this should not impact anything at runtime. If @garyrightbot would like to test to validate since they wanted this, that would be great.


amakarow:
Thank you. I merged this.

PR #238, opened 2020-10-23
Updateable robot model

nickvaras:
This allows the robot footprint model to be updated. The current "Is_footprint_dynamic" dynamically reconfigurable parameter only updates the global planner footprint to check feasibility. This change allows updating the actual model, even changing the model type.  To use, first update the values on the parameter server, then call the dynamic reconfigure service. No need to reconfigure anything though, it's the call that triggers the update. So if the footprint_dynamic is set to true, calling the service to set it to true again will update the model  (unless someone wants to implement a separate service call...).

Tested on melodic.


amakarow:
Thanks.

PR #243, opened 2020-11-07
add inline and override tags to the vertices

dorezyuk:
Hi guys,
A small pr cleaning up the vertex types
- add inline and override tags
- remove default destructors
- remove outdated documentation
- remove ros/console include
- fix missing white space in the ostream


amakarow:
Thanks

jcmonteiro:
Hey there,

I know this has been merged already, but with regards to the `inline` keyword, it is somewhat deprecated already in the context of the changes in this PR (suggesting that the compiler should repeat the function body). The compiler will happily ignore the suggestion and make its own choices. It is worth taking this into account for future PRs.

PR #244, opened 2020-11-07
bulk-delete the edges registered at the vertices

dorezyuk:
PR simplifies the formulation of the base-edges and offers a small speed-up of the clearGraph function.

1. The constructors can be removed, since g2o already initializes the vertices members to nullptr
- https://github.com/RainerKuemmerle/g2o/blob/832b96340c626ed75b767ec4867b95ce7abf8093/g2o/core/base_unary_edge.h#L55
- https://github.com/RainerKuemmerle/g2o/blob/832b96340c626ed75b767ec4867b95ce7abf8093/g2o/core/base_binary_edge.h#L69

2. The deletion of the non-owning _vertices.edges-member can be done at once for all edges inside the clearGraph function.
In this way don't have to pay the price for the lookup inside std::set<Edge>::erase

### DIFF
--- a/include/teb_local_planner/g2o_types/base_teb_edges.h
+++ b/include/teb_local_planner/g2o_types/base_teb_edges.h
@@ -75,26 +75,6 @@ public:
   using g2o::BaseUnaryEdge<D, E, VertexXi>::computeError;
     
   /**
-   * @brief Construct edge.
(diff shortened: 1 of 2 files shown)


amakarow:
Thanks.

PR #247, opened 2020-11-09
Dima/cleanup

dorezyuk:
PR cleans up the code a little bit:
1. remove unused getError methods and the definition imported by it (I suppose this was used some time ago and left here)
2. remove the overwrite of resize, since g2o sets the all vertices already to 0. see 
https://github.com/RainerKuemmerle/g2o/blob/c64f79af4b7308a0505a05831f46e0dffb9a0c45/g2o/core/hyper_graph.cpp#L70
3. add a config-interface, so we can reuse it in all base-edge classes (reduces the copy and paste code)

### DIFF
--- a/include/teb_local_planner/g2o_types/base_teb_edges.h
+++ b/include/teb_local_planner/g2o_types/base_teb_edges.h
@@ -54,8 +54,28 @@
 
 namespace teb_local_planner
 {
-    
-    
+
+/**
+ * @brief Common interface for setting/getting the TebConfig.
+ *
+ * Class offers a setter/getter interface used by the edges of this library.
+ */
+struct ConfigInterface{
+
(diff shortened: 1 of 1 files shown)


include/teb_local_planner/g2o_types/base_teb_edges.h
```diff
@@ -54,8 +54,28 @@
...
+
+/**
+ * @brief Common interface for setting/getting the TebConfig.
+ *
+ * Class offers a setter/getter interface used by the edges of this library.
+ */
+struct ConfigInterface{
+
+  inline const TebConfig* getTebConfig() const noexcept
+  {
```
dorezyuk, line 66: getter actually not required, so it could be dropped

PR #248, opened 2020-11-10
Handle empty polygon in customObstacle callback

DarthBubi:
This prevents the planner from crashing when publishing an empty polygon in an ObstacleMsg to the `test_optim_node`.


amakarow:
Thanks, merged.

PR #253, opened 2020-11-20
sync with changes from nav2

jediofgever:
A few changes in `nav2_core::Controller` base class causes `teb_local_planner` not to build in the current form, this PR fixes it. It is some minor pointer compatibility issue

### DIFF
--- a/teb_local_planner/include/teb_local_planner/teb_local_planner_ros.h
+++ b/teb_local_planner/include/teb_local_planner/teb_local_planner_ros.h
@@ -106,7 +106,7 @@ public:
    * @param costmap_ros Cost map representing occupied and free space
    */
   void configure(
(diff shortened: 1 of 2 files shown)


p0rys:
The changes seems to work only on the main branch of navigation2 but not the version that can be installed on foxy via apt. I tried and it failed compiling.

Maybe it should be splitted in two parts. One with needed foxy changes into foxy-devel (removing the "override" in header file) and the other one with changes for navigation2 main branch (SharedPtr to WeakPtr).

jediofgever:
Thats right , I think it is better to have an `main` or `ros2` branch that tracks upstream updates from ros2.  people who use foxy distro should not  require changes in this PR, I targeted foxy-devel cause there is no main ros2 branch

Issue #245, opened 2020-11-09
Segfault on dynamic-reconfigure with new UpdateRobotModel Patch


dorezyuk:
Hey guys,

with the new update robot model patch, we have a segfault if we run dynamic reconfigure without clearing the graph.

### Reason

Every obstacle edge holds a raw pointer to the object managed by robot_model_. If we update that robot_model_, this raw-pointer gets invalidated. If we don't run clearGraph, then the edges try to a access invalid memory and we segfault.

### Reproduction

Give the robot a goal, and trigger dynamic reconfigure (in a loop, since its hard to control when your clearGraph gets called) repetitively.

PR #255, opened 2020-11-23
fix segfault at robot-model update

dorezyuk:
PR fixes issue #245 

The fix moves the ownership of the footprint from the optimizers to the teb_config class. While this change is debatable (keeping stuff separate is cool) it allows a straight forward fix of the tagged issue and makes dedicated interfaces for setting the footprint obsolete. It might also be acceptable to view the footprint as part of the config for a robot controller (but as said, also fine not to do so).

The positive side effects of this fix are
* the footprint is then constant for the optimal_planner and homotopy_class_planner
* the "default footprint" (PointRobotFootprint) is defined only once.
* one could make the footprint unique_ptr instead of shared, making the ownership clearer

### DIFF
--- a/include/teb_local_planner/g2o_types/edge_dynamic_obstacle.h
+++ b/include/teb_local_planner/g2o_types/edge_dynamic_obstacle.h
@@ -80,7 +80,7 @@ public:
   }
   
   /**
-   * @brief Construct edge and specify the time for its associated pose (neccessary for computeError).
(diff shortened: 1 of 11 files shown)


amakarow:
The positive side effects are clearly nice. However, as you already pointed out, I am still not hundert percent sure if we should really move the robot model to the config class ;).

corot:
Equivalent change applied on #393

Issue #257, opened 2020-12-03
when robot navigates close to wall or obstacle, planner publish large amout of teb_poses


charon-cheung:
I upload the [video](https://www.youtube.com/watch?v=ETORdzBcT0w), this rarely happens, I found the global path changed in the end, is that the reason ?

jcmonteiro:
It could be that this is caused by the global path changing to an unfeasible final pose. The first and the final poses are fixed vertices in the optimization.

It looks like the algorithm had converged well from your video, but at the very end, your final pose changed to an unfeasible one. You could try to reproduce in two ways:

- Change the final pose at the very end to an unfeasible one.
- Change the final pose at the very end to another feasible one.

Both of the above roughly at the same final position (just like in your video).

charon-cheung:
> It could be that this is caused by the global path changing to an unfeasible final pose. The first and the final poses are fixed vertices in the optimization.
> 
> It looks like the algorithm had converged well from your video, but at the very end, your final pose changed to an unfeasible one. You could try to reproduce in two ways:
> 
> * Change the final pose at the very end to an unfeasible one.
> * Change the final pose at the very end to another feasible one.
> 
> Both of the above roughly at the same final position (just like in your video).

I thought you are right, but I'm using GlobalPlanner::makePlan, it doesn't change final position, I think you are talking about Navfn::makePlan, which will do. So I don't think that's the reason

Noah-Lau:
I got a similar problem, have you figure it out?

Issue #200, opened 2020-03-26
velocity command disappears for a while


zengxiaolei:
Hi, Christoph,
I use TEB-local-planner in ROS melodic (0.8.4), Ubuntu 18.04. 
As shown in Fig. 1, velocity command from local planner disappears for a while at time 2260.5s and  at 2264.3s, and velocity frequency is relative low after these time-points for some time, which results in strange local plan: the local path can form one or more circles.

![teb-vel-plot](https://user-images.githubusercontent.com/34747839/77624262-81a62a80-6f7c-11ea-9fe7-6612af79ca5c.png)

Fig. 1
move_base_vel : original velocity topic of move_base
cmd_vel : smoothed move_base_vel
odom: odom velocity feedback

This phenomenon can happens, especially at start point or end point. I don't know if this is a problem caused by inappropriate parameters or the algorithm itself. I look forward to your help.

adamkrawczyk:
I think I have a similar problem when robot moves sometimes planner stops sending cmd_vel data and robot uses last published velocity. When this happens I can see warnings that `control loop missed desired frequency`. I was wondering if this can be caused by not enough computation power but I'm using 16cpu and usage is about 50%. 

ROS Kinetic

srsidd:
I'm seeing a similar problem, except I never see the planner recover.. Were you able to figure this out?

adamkrawczyk:
Unfortunately I haven't and changed to other local planner.

zengxiaolei:
Hi, I've solved the problem. In my case,  the reason is that the global plan path is not smooth. After fix global path parameter, the problem is solved.

Issue #61, opened 2017-12-06
Detect teb convergence problem


NeryHenrique:
Hi Christoph,

Sometimes, due a different number of factors, teb cannot converge the path to a good solution.
One very important thing, since in these cases the robot can go anywhere... is to detect that case and stop the cmd vel, like in the unfeasible plan. 
I m thinking in analyse the poses generated (eg: if the poses are too distant from each other or a measure of how the angle between multiple poses were selected. 
Maybe we can do it in the answer of G2o or in the cost step? 
What to do think about this functionality?

Thanks for the awesome work!

croesmann:
Hi @NeryHenrique,

thank you for your feedback! I am struggling with this question for quite a while. One option is to verify if the constraints of the optimization problem are satisfied (which are included only as penalties in the optimization problem). However, often the problem seems to be "infeasible" at an early stage, but after some more iterations, the solution reveals feasibility. One could try to also check the *relative* convergence based on a cost tolerance *and* additionally the constraints to ensure constraint checking only in a converged state. The question is, how robust is the selection of the tolerance threshold. And this solution does not take oscillating solutions (costs) into account. But it could be worth a try ;-)

The current implementation trys to check only the collision constraints on a first subset of the planned trajectory (parameter `feasibility_check_no_poses`), however, this is still not perfect.
Another attempt to detect planner failures is introduced in b1ffbb5bf5d24207e27e09496a48110833a3cceb in which I defined a FailureDetector class which analyzes the control actions only, e.g., in order to detect oscillations. However, this does not include checking the complete plan. 
If you have some ideas or time to test such approaches, please let me know and I would be happy to merge them into the code.

Cheers,
Christoph

PR #259, opened 2020-12-08
[feature] Divergence detection

jcmonteiro:
### Why
This is an edge case in my application, but if an obstacle suddenly appears in front of the robot, TEB's solution might diverge quite noticeably. As an example, I obtained the image below by suddenly dropping an obstacle in front of the robot in simulation.

image

### Summary
After some investigation, I've observed that (given my configuration) the value of `chi2` in the optimizer statistics remains around 1 when the robot is navigating. Still, it increases drastically if the trajectory diverges, reaching values above 1000. This seemed to me to be the best parameter to monitor to detect divergence, but I might be missing the context of other applications (my robot is a differential drive one, which runs at most at 0.75 m/s).

### Related Issues
Might help with #61 #257 #7 #200


amakarow:
This looks nice, thanks for this PR.

PR #263, opened 2020-12-16
Timestamp fix alongwith addition of a dt_force_equal parameter

BhaktiDanve:
Hello Sir,

Our Professorship has been using teb local planner for a long time and indeed it has helped solve many of our trajectory related problems. While visualising some of our test trajectories in Rviz generated by teb local planner, there occurs an unequal distribution of time-differences between the poses as shown in the figure below:

dt__ex2a

After looking into the code, it was observed that the 'autoResize of poses' logic is executed before calling the Graph Optimiser (See [optimal_planner.cpp](../../teb_local_planner/tree/melodic-devel/src/optimal_planner.cpp)) The current logic for autoResize (See [timed_elastic_band.cpp](../../teb_local_planner/tree/melodic-devel/src/timed_elastic_band.cpp)) creates new poses by interpolation based on comparing current-time-difference value to desired-time-difference(dt_ref) +/- some hysteresis(dt_hysteresis). But this might lead to oscillations in the number of poses generated in each cycle resulting in unequally distributed poses. The problem is further deteriorated if the default fastmode is used. The oscillations are visualised with the help of the following graph:

schritt_eins

Table: time-stamps of sequential poses
Graph: X-axis: no. of rep-cycles, Y-axis: no of poses in each rep-cycle

As an improvement to the current 'autoResize' logic to avoid the above problems, 
- an additional parameter 'dt_force_equal' is added
- If the current-time-difference is greater than desired-time-difference(dt_ref) +/- some hysteresis(dt_hysteresis) and dt_force_equal is true:
  - only if the current-time-difference is greater than twice the desired-time-difference(dt_ref):
        -> a new pose is added between adjacent poses
        -> current-time-difference is re-compared with the desired-time-difference(dt_ref) in the next iteration
  - else, the current-time-difference is set to the desired value and the extra time is pushed to the next time-difference, if possible. This step might create some more tension on 
        the Graph optimizer, but as expected, the optimizer correctly updates the pose values based on the updated time-difference.
- Else if the current-time-difference is less than desired-time-difference(dt_ref) +/- some hysteresis(dt_hysteresis) and dt_force_equal is true:
  - the logic remains the same, but current-time-difference(which is not the last one) is re-compared with the desired-time-difference(dt_ref) in the next iteration

After making the above modifications in the code, it results in equal distribution of time-differences between the poses as shown in the figure below:

dt__ex2b

I hope these changes will be included to the current algorithm.

### DIFF
--- a/include/teb_local_planner/timed_elastic_band.h
+++ b/include/teb_local_planner/timed_elastic_band.h
@@ -503,8 +503,9 @@ public:
    * @param max_samples maximum number of samples that should not be exceeded during resizing
    * @param fast_mode if true, the trajectory is iterated once to insert or erase points; if false the trajectory
    *                  is repeatedly iterated until no poses are added or removed anymore
+   * @param force_equal if true, the trajectory is forced to use the correct time diffs
    */    
-  void autoResize(double dt_ref, double dt_hysteresis, int min_samples = 3, int max_samples=1000, bool fast_mode=false);
+  void autoResize(double dt_ref, double dt_hysteresis, int min_samples = 3, int max_samples=1000, bool fast_mode=false, bool dt_force_equal=false);
 
   /**
    * @brief Set a pose vertex at pos \c index of the pose sequence to be fixed or unfixed during optimization.
--- a/include/teb_local_planner/timed_elastic_band.h
+++ b/include/teb_local_planner/timed_elastic_band.h
@@ -505,7 +505,7 @@ public:
    *                  is repeatedly iterated until no poses are added or removed anymore
    * @param force_equal if true, the trajectory is forced to use the correct time diffs
    */    
-  void autoResize(double dt_ref, double dt_hysteresis, int min_samples = 3, int max_samples=1000, bool fast_mode=false, bool dt_force_equal=false);
+  void autoResize(double dt_ref, double dt_hysteresis, int min_samples = 3, int max_samples=1000, bool fast_mode=false);
 
   /**
    * @brief Set a pose vertex at pos \c index of the pose sequence to be fixed or unfixed during optimization.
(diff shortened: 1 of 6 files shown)


amakarow:
Thanks for this PR. This looks good. Just to avoid having too many parameters, do you think it is safe to replace the old logic with your new one without introducing the new parameter dt_force_equal?

BhaktiDanve:
> Thanks for this PR. This looks good. Just to avoid having too many parameters, do you think it is safe to replace the old logic with your new one without introducing the new parameter dt_force_equal?

Hello @amakarow,

Thanks for your response. Of course, we can eliminate the parameter and replace the old logic with the new one. The new logic is well tested at our part. Parameter was simply created to provide this logic as an alternative.

Once I receive your confirmation, I will update this pull request without the parameter and with the replaced logic.


Thanks and Regards,
Bhakti Danve

BhaktiDanve:
Hallo @amakarow ,

I have now updated the pull request without the parameter dt_force_equal, to avoid too many parameters.

Regards,
Bhakti

include/teb_local_planner/timed_elastic_band.h
```diff
@@ -503,6 +503,7 @@ class TimedElasticBand
    * @param max_samples maximum number of samples that should not be exceeded during resizing
    * @param fast_mode if true, the trajectory is iterated once to insert or erase points; if false the trajectory
    *                  is repeatedly iterated until no poses are added or removed anymore
+   * @param force_equal if true, the trajectory is forced to use the correct time diffs
```
amakarow: Could you please remove this parameter description? Unfortunately, I do not have write permissions in your branch, so I was not able to take over.

BhaktiDanve: Hallo @amakarow ,

Thanks for your response. The parameter description is now removed.

peterweissig:
Hello,

it has been a while since the last activity ... is there anything else you need from us, before you are able to merge our request ?

Thank you & stay healthy
  Peter

amakarow:
No this is perferkt right now, thank you! Merged.

PR #265, opened 2020-12-20
When transforming the path, use robot pose to choose the right leg on crossings

corot:
This prevents to alternate between two options when there are loops in the path, e.g. in this video:

https://user-images.githubusercontent.com/322610/102725935-df548100-431a-11eb-8075-563c313dc1d6.mp4


amakarow:
Nice, thanks. Unfortunately, the video does not work. However, this looks good, merged.

corot:
> Nice, thanks. Unfortunately, the video does not work. However, this looks good, merged.

Thanks. Do you get any error msg when clicking in the video?  I'm curious about why is not working

PR #267, opened 2020-12-29
Foxy devel

ByLiZhao:
A foxy-devel branch is forked from the "ros2-master" branch of the "teb_local_planner" project.
Building errors are fixed when building against the foxy-devel branch of Navigation2 project.


amakarow:
Big thanks for fixing foxy-devel branch. I squashed and merged this.

PR #276, opened 2021-02-15
Wait for robot to stop before setting goal reached

ayushgaud:
### Feature addition ###
Planners like [dwa_local_planner](http://wiki.ros.org/dwa_local_planner) and [base_local_planner](http://wiki.ros.org/base_local_planner) uses [base_local_planner::LocalPlannerLimits](http://docs.ros.org/en/api/base_local_planner/html/classbase__local__planner_1_1LocalPlannerLimits.html) class and enables users to set stop velocities. This is used to make sure that the robot has actually stopped before returning success. This avoids issues of overshooting the goal after the planner has returned success.

#### Changes ####
In this PR two new parameters are introduced, `trans_stopped_vel` and `theta_stopped_vel` that ensures the robot achieves translational and rotational velocity below specified thresholds, respectively, before returning success.

#### Testing ####
Tested by manually publishing mock data to the odom topic.

### DIFF
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -155,12 +155,13 @@ grp_goal.add("free_goal_vel",   bool_t,   0,
 	False)
 
 grp_goal.add("trans_stopped_vel",   double_t,   0,
-       "Below what maximum velocity we consider the robot to be stopped in translation",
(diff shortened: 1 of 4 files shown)


corot (approved):
Tested with simulated turtlebot; works as expected

amakarow:
Big thanks, merged.

PR #280, opened 2021-03-10
[ros2-master] Updated parameters for ros2 master sync

### LINKED ISSUE #220: [ROS2] Update teb_params.yaml in order to properly run tb3 simulation
OS: Ubuntu 20.04
Dist: ROS2 Foxy Binaries
Nav2: 69977cd83d9040df3422d8a2e564715b6002f3fb
TEB: 8bd805e29ef629a003880e22d67d9529da76a87d

The current teb_params.yaml in the ros2-master branch does not allow the tb3 simulation to successfully run by preventing the controller node from progressing to an active state. 

### Steps to replicate:
Build the teb_local_planner package for ROS2 Foxy and run `teb_tb3_simulation_launch.py`.

### Update:
I have tried replacing the current parameters with the ones found in issue https://github.com/rst-tu-dortmund/teb_local_planner/issues/203 with little luck, although it does seem to allow the nav2 stack to startup fine.  However, all I get with that configuration is the following output while the tb3 just rotates in place.

```
[bt_navigator-11] [INFO] [1595707851.384197642] [bt_navigator]: Begin navigating from current location to (0.65, 1.84)
[controller_server-8] [INFO] [1595707851.418366655] [controller_server]: Received a goal, begin computing control effort.
[controller_server-8] [INFO] [1595707852.468641585] [controller_server]: Passing new path to controller.
[controller_server-8] [INFO] [1595707853.518666932] [controller_server]: Passing new path to controller.
[controller_server-8] [INFO] [1595707854.518639488] [controller_server]: Passing new path to controller.
[controller_server-8] [INFO] [1595707855.568666246] [controller_server]: Passing new path to controller.
```

AlexKaravaev:
Resolves #220.
With this params I was able to launch nav2 branch master with teb_local_planner as controller plugin


AlexKaravaev:
@SteveMacenski Should I remove params, not related to controller_server from teb_params.yaml?
I've added them, because the ones that are present now will not work with current nav2 master

SteveMacenski:
Yeah, lets just delete all the launch files and use only the config file section for controller server. I don't want to maintain 2 sets of identical parameters for everything else. If someone's swapping out controllers, they can just copy over the parameters.

Thanks for this!

AlexKaravaev:
Makes sense, I've deleted launch files and left only controller server related parameters

teb_local_planner/launch/teb_params.yaml
```diff
@@ -1,213 +1,112 @@
...
+      roadmap_graph_area_width: 5.0
+      roadmap_graph_area_length_scale: 1.0
+      h_signature_prescaler: 0.5
+      h_signature_threshold: 0.1
+      obstacle_heading_threshold: 0.45
+      switching_blocking_period: 0.0
+      viapoints_all_candidates: True
+      delete_detours_backwards: True
+      max_ratio_detours_duration_best_duration: 3.0
+      visualize_hc_graph: False
```
SteveMacenski: The directory should be renamed `params` or `config` but otherwise LGTM!

SteveMacenski (approved):

SteveMacenski:
Thanks!

doisyg:
This PR breaks compilation!
I believe it is due to the complete deletion of the launch directory and the fact that it is still declared in an install step here:
https://github.com/rst-tu-dortmund/teb_local_planner/blob/a805dddc4773542f89332b972c6b570b3f464177/teb_local_planner/CMakeLists.txt#L139

PR #281, opened 2021-03-11
Fix ros2-master

doisyg:
fix compilation error due to https://github.com/rst-tu-dortmund/teb_local_planner/pull/280

### DIFF
--- a/teb_local_planner/CMakeLists.txt
+++ b/teb_local_planner/CMakeLists.txt
@@ -136,7 +136,7 @@ install(FILES
 
 install(PROGRAMS scripts/cmd_vel_to_ackermann_drive.py DESTINATION lib/${PROJECT_NAME})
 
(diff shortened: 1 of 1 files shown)


AlexKaravaev:
Yes, my mistake. I forgot to test compilation after changing directory name. But maybe we should not just comment out this line
```
install(DIRECTORY launch DESTINATION share/${PROJECT_NAME})
```
But instead just change it to
```
install(DIRECTORY params DESTINATION share/${PROJECT_NAME})
```

doisyg:
Changes for `install(DIRECTORY params DESTINATION share/${PROJECT_NAME})` in last commit

doisyg:
Redundant with (included in) : https://github.com/rst-tu-dortmund/teb_local_planner/pull/282

Issue #283, opened 2021-03-11
[ros2-master] Maintenance and sync with ROS1 branches


doisyg:
Hello TEB planners,
After discussion with @SteveMacenski, me and me team will help syncing and putting up to date the `ros2-master` branch (and targeting a release for galactic).
It looks like the last sync with `melodic-devel` has been done one year ago (21/02/2020), furthermore, it doesn't look like `melodic-devel` and `noetic-devel` are in sync after the 17/10/2020.
The fact that the directory structure is not the same between the ROS1 branches and `ros2-master` is not helping with keeping them in sync.

@BriceRenaudeau did the job of tracking the differences between the 3 branches below : 
| **Source** date |`melodic-devel` | `noetic-devel` | `ros2-master` |
| --- |--- |--- | --- |
**02/05/2021**| | | |
| |**Merge pull request #263 from TUC-ProAut/melodic-local** https://github.com/rst-tu-dortmund/teb_local_planner/commit/073e4c52ccf2ef3679c71a98b03cf8f0bec93e76| --- | |
**21/03/2021**| | | |
| |**Parameter Description for force_equal removed** https://github.com/rst-tu-dortmund/teb_local_planner/commit/33253193610c7784808def198a9652f2f6f6f1e6| --- | |
**08/03/2021**| | | |
| |**Merge  robot stopped before goal reach [#276](https://github.com/rst-tu-dortmund/teb_local_planner/pull/276)**| --- | SYNCED with https://github.com/rst-tu-dortmund/teb_local_planner/pull/284 |
| |**Fix free_goal_vel [#250](https://github.com/rst-tu-dortmund/teb_local_planner/pull/250)**|--- | SYNCED with https://github.com/rst-tu-dortmund/teb_local_planner/pull/284 |
**04/01/2021**| | | |
| |**Merge use left leg on cross road [#265](https://github.com/rst-tu-dortmund/teb_local_planner/pull/265)**| --- | SYNCED with https://github.com/rst-tu-dortmund/teb_local_planner/pull/284 |
| |**Merge divergence detection [#259](https://github.com/rst-tu-dortmund/teb_local_planner/pull/259)**| --- | SYNCED with https://github.com/rst-tu-dortmund/teb_local_planner/pull/284 |
[...]

PR #284, opened 2021-03-14
[ros2-mater] Sync with melodic-devel 

doisyg:
As discussed here: https://github.com/rst-tu-dortmund/teb_local_planner/issues/283

One main point to discuss before merging: goal reach checking. There was some melodic code that was not ported : 
https://github.com/rst-tu-dortmund/teb_local_planner/blob/0d242c0aff4f81b4ff5d67d7e4f970a8f454ad4a/src/teb_local_planner_ros.cpp#L287-L301
In the current ros2 branch there is this code marked for goal reach checking : 
https://github.com/rst-tu-dortmund/teb_local_planner/blob/e3acd7d0d7519406cd136a6d8fead0f783025363/teb_local_planner/src/teb_local_planner_ros.cpp#L299-L302
But I don't get how it is done.

More generally, maybe we should remove the code related to goal checking since in nav2 it is done by a controller pluggin, @SteveMacenski, what do you think ?


SteveMacenski:
@doisyg I changed the base of your PR to ros2, I assume that's what you meant :-)

SteveMacenski (approved):
From a high level, seems reasonable. At this junction, would testing be wise? Or after this goes in?

SteveMacenski:
>  But I don't get how it is done.

I don't think it is done. `global_goal` is never used to compare against. That looks like a block of dead code

doisyg:
> @doisyg I changed the base of your PR to ros2, I assume that's what you meant :-)

Yes, thanks!

doisyg:
> > But I don't get how it is done.
> 
> I don't think it is done. `global_goal` is never used to compare against. That looks like a block of dead code

Agreed, I added a commit which removes the parameters which were used in this goal reach check and which are not used anywhere in the ROS2 code : `trans_stopped_vel`, `theta_stopped_vel`, `yaw_goal_tolerance` and `complete_global_plan`. I believe they are functionaly replaced in ROS2 by the goal checker (hence my recent PR https://github.com/ros-planning/navigation2/pull/2248)
There is still `xy_goal_tolerance` which is used here:
https://github.com/rst-tu-dortmund/teb_local_planner/blob/e3acd7d0d7519406cd136a6d8fead0f783025363/teb_local_planner/src/graph_search.cpp#L102
and here:
https://github.com/rst-tu-dortmund/teb_local_planner/blob/e3acd7d0d7519406cd136a6d8fead0f783025363/teb_local_planner/src/graph_search.cpp#L233
and that we should, I believe, get from the `goal_checker` param (and if possible, dynamically), but this will be the task of another PR

SteveMacenski:
>  and that we should, I believe, get from the goal_checker param (and if possible, dynamically), but this will be the task of another PR

Indeed, if fact this wouldn't build with Nav2 without adding that `goal_checker` arg after https://github.com/ros-planning/navigation2/pull/2247 is merged. You'll have access to it immediately after that to remove that last param

SteveMacenski:
Merge this? What about testing?

doisyg:
> > and that we should, I believe, get from the goal_checker param (and if possible, dynamically), but this will be the task of another PR
> 
> Indeed, if fact this wouldn't build with Nav2 without adding that `goal_checker` arg after [ros-planning/navigation2#2247](https://github.com/ros-planning/navigation2/pull/2247) is merged. You'll have access to it immediately after that to remove that last param

Ok, I will keep an eye on this

doisyg:
> Merge this? What about testing?

Yes please, building against last nav2 main. Basic test with a real robot done.
Still planning to do extensive testing of each feature once everything is merged. I updated the plan here: https://github.com/rst-tu-dortmund/teb_local_planner/issues/283 (end of first post)

PR #288, opened 2021-03-18
[ros2-master] fix isTrajectoryFeasible

doisyg:
`isTrajectoryFeasible` was broken. It was using `dwb_critics::ObstacleFootprintCritic::scorePose` for pose validity evaluation, however this function is throwing exceptions, notably when evaluating poses touching lethal cells (@SteveMacenski, is this by design ?)
https://github.com/ros-planning/navigation2/blob/3bf1f7a93d740c54b4c7ef7fc6dd914175bf46be/nav2_dwb_controller/dwb_critics/src/obstacle_footprint.cpp#L154-L159
Added catch of these exceptions and fixed the visualization of the invalid poses


SteveMacenski:
@doisyg that is DWB's design, it throws exceptions when in collision to basically back out to the trajectory being scored to move onto the next one.

Wouldn't be _my_ design, but it is intentional

SteveMacenski (approved):

PR #295, opened 2021-04-13
robot footprint model access specifiers loosened

rayvburn:
I propose minor changes in `robot_footprint_model` to allow more extension possibilities (for e.g. planners which use similar environment representation).

- `BaseRobotFootprint` class
    - `getInscribedRadius` was a non-const method even though none of the members are changed in any of derived classes
- `TwoCirclesRobotFootprint` class
    - `front_offset`, `front_radius`, `rear_offset`, `rear_radius` are marked as private which kinda blocks extension possibilities
- `LineRobotFootprint` class
    - `transformToWorld` method was private, similarly, `line_start` and `line_end`

### DIFF
--- a/include/teb_local_planner/robot_footprint_model.h
+++ b/include/teb_local_planner/robot_footprint_model.h
@@ -107,7 +107,7 @@ public:
    * @brief Compute the inscribed radius of the footprint model
    * @return inscribed radius
    */
-  virtual double getInscribedRadius() = 0;
+  virtual double getInscribedRadius() const = 0;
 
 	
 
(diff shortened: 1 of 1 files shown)


amakarow:
Thank you, looks good, but kinetic is outdated. Do you really need this in kinetic? We would prefer to merge into melodic/noetic.

rayvburn:
Sure that Kinetic is pretty old but e.g. PAL's software for [`TiAGo robot`](https://pal-robotics.com/robots/tiago/) seems to work best with that distro. Having that in Kinetic would be very helpful!

Issue #10, opened 2016-06-08
Allow setting of minimal velocity


nlimpert:
Hi,

is it possible to implement a minimal velocity? We are using the teb_local_planner with success on either simulated or real ackermann driven robots.
But in some cases we are facing the problem that the robot base cannot execute the desired velocity.
E.g. it would only start driving at a minimal velocity of 0.1 m/s.

Best regards and thumbs up for the great work!

croesmann:
Hi @balamesh,

in general, adding a lower bound on the minimum velocity should be possible. However, integrating this into the optimization scheme significantly increases complexity since we already support switching between forward and backwards motions  (it might introduce non-smooth cost functions or even more local minima).

In my opinion, adjusting the controls/velocities _after_ obtaining the optimization result should work fine in many applications, e.g. using a simple three-point-controller before forwarding the velocities to the actual base controller:

```
if ( abs(v_out >= v_min) ) v_out = v;
else if ( abs(v_out >= eps) ) v_out = sign(v)*v_min; // assuming symmetric friction properties
else v_out = 0; 
```

However, the threshold `eps` depends on
- the maximum acceleration (the robot must be able to reach the subsequent discrete state of the local plan) 
- and the goal tolerance (the robot must be able to reach the desired goal region)

If `v_min` is small and `eps` is chosen appropriately, it should not influence stability significantly to my mind.

It would be really straight forward to add those few lines to the `computeVelocityCommand()` method in `teb_local_planner_ros.cpp` but in my opinion it is more suitable and intuitive if these platform dependent heuristics are included in the robot's base controller as long it is not subject to the actual optimization/planning scheme.

nlimpert:
Yes thanks a lot for your help. We managed to solve this by implementing a minimum velocity on the controller side.
And sorry for the late response!

croesmann:
For the record, since it might be related:
Configuring the acceleration limits carefully is crucial as well.
Since the controller operates time-optimally, changing the maximum acceleration (in combination with dt_ref) should result in higher initial velocities.

PR #322, opened 2021-10-08
control_look_ahead_poses can now be dynamically reconfigured

siferati:


corot:
We have noticed that this parameter is very useful for robots a bit lethargic reacting to commands. Setting to a higher than 1 value reduces progressively oscillations around the global path. However, 1 is the perfect value for precise control of the same lethargic robots when moving slow.

AlperenKosem:
for instance :
[https://github.com/rst-tu-dortmund/teb_local_planner/issues/10](url)

SteveMacenski (approved):

SteveMacenski:
@corot I'm not an admin on here on the ROS 1 side, but for something trivial like this I can merge it, but that shouldn't be the baseline expectation.

corot:
> @corot I'm not an admin on here on the ROS 1 side, but for something trivial like this I can merge it, but that shouldn't be the baseline expectation.

Got it. I volunteered myself to help maintaining for ROS1, so if you can provide some reviews, it will be more than enough.
It's of if I keep tagging you for future reviews? Thanks!

SteveMacenski:
I cannot provide reviews for ROS 1, I have too many other responsibilities, I cannot add this among them

PR #329, opened 2021-11-23
Added Minimum Translational Velocity Limit

AlperenKosem:
Most planners have no minimum translational velocity limit and, People who want to add min limit solve this problem in low level controller.  "min_vel_x" parameter is added to rqt_reconfiguration and configuration file.


AlperenKosem:
for instance : 

-  #10

include/teb_local_planner/teb_config.h
```diff
@@ -104,6 +104,7 @@ class TebConfig
     bool is_footprint_dynamic; //<! If true, updated the footprint before checking trajectory feasibility
     bool use_proportional_saturation; //<! If true, reduce all twists components (linear x and y, and angular z) proportionally if any exceed its corresponding bounds, instead of saturating each one individually
     double transform_tolerance = 0.5; //<! Tolerance when querying the TF Tree for a transformation (seconds)
+    double min_vel_x;
```
corot: move together with the other velocities (e.g. after max_vel_x), as on .cfg and add the same comment here

src/teb_config.cpp
```diff
@@ -83,6 +83,8 @@ void TebConfig::loadRosParamFromNodeHandle(const ros::NodeHandle& nh)
   nh.param("is_footprint_dynamic", robot.is_footprint_dynamic, robot.is_footprint_dynamic);
   nh.param("use_proportional_saturation", robot.use_proportional_saturation, robot.use_proportional_saturation);
   nh.param("transform_tolerance", robot.transform_tolerance, robot.transform_tolerance);
+  nh.param("min_vel_x", robot.min_vel_x, robot.min_vel_x);
```
corot: move together with the other velocities (e.g. after max_vel_x), as on .cfg

cfg/planner.yaml
```diff
@@ -0,0 +1,113 @@
+# ########### Teb Local Planner
```
corot: this is out of scope, isn't it?

cfg/TebLocalPlannerReconfigure.cfg
```diff
@@ -86,7 +86,11 @@ grp_robot.add("max_vel_x", double_t, 0,
 
 grp_robot.add("max_vel_x_backwards", double_t, 0, 
 	"Maximum translational velocity of the robot for driving backwards",
-	0.2, 0.01, 100)   
+	0.2, 0.01, 100)  
+
+grp_robot.add("min_vel_x", double_t, 0,
```
corot, line 95: makes a bit more sense right after max_vel_x, imho

corot (changes requested):
nice addition; I guest it can be useful on some robots with picky controllers

can you move the new param together with the other velocities (e.g. after max_vel_x), as on .cfg?
Also, there're a bunch of spurious line breaks and a configuration file out of scope

AlperenKosem:
thanks @corot !

Purious line breaks and a configuration file are deleted, and location of min_vel_x parameter was edited.

cfg/TebLocalPlannerReconfigure.cfg
```diff
@@ -82,11 +82,15 @@ grp_robot = gen.add_group("Robot", type="tab")
 
 grp_robot.add("max_vel_x", double_t, 0, 
 	"Maximum translational velocity of the robot",
-	0.4, 0.01, 100)   
+	0.4, 0.01, 100)
+
+grp_robot.add("min_vel_x", double_t, 0, 
+	"Minimum translational velocity of the robot",
+	0.01, 0.01, 100)
```
corot: I think 0 is a better default, so this PR won't alter current behavior

include/teb_local_planner/teb_local_planner_ros.h
```diff
@@ -355,7 +355,7 @@ class TebLocalPlannerROS : public nav_core::BaseLocalPlanner, public mbf_costmap
    * @param max_vel_x_backwards Maximum translational velocity for backwards driving
    */
   void saturateVelocity(double& vx, double& vy, double& omega, double max_vel_x, double max_vel_y,
-                        double max_vel_theta, double max_vel_x_backwards) const;
+                        double max_vel_theta, double max_vel_x_backwards, double min_vel_x) const;
```
corot: here too, makes sense after max_vel_x

src/teb_config.cpp
```diff
@@ -83,6 +84,8 @@ void TebConfig::loadRosParamFromNodeHandle(const ros::NodeHandle& nh)
   nh.param("is_footprint_dynamic", robot.is_footprint_dynamic, robot.is_footprint_dynamic);
   nh.param("use_proportional_saturation", robot.use_proportional_saturation, robot.use_proportional_saturation);
   nh.param("transform_tolerance", robot.transform_tolerance, robot.transform_tolerance);
+  
+
```
corot: ?

include/teb_local_planner/teb_local_planner_ros.h
corot: same for all saturateVelocity methods

src/teb_local_planner_ros.cpp
```diff
@@ -869,13 +869,28 @@ double TebLocalPlannerROS::estimateLocalGoalOrientation(const std::vector<geomet
...
 {
-  double ratio_x = 1, ratio_omega = 1, ratio_y = 1;
-  // Limit translational velocity for forward driving
-  if (vx > max_vel_x)
-    ratio_x = max_vel_x / vx;
-  
+  double ratio_x = 1, ratio_omega = 1, ratio_y = 1, ratio_x_min = 1;
+
+  // backward driving
+  if(vx < 0)
```
corot:
```suggestion
  if (vx < 0)
```

src/teb_local_planner_ros.cpp
```diff
@@ -901,9 +916,10 @@ void TebLocalPlannerROS::saturateVelocity(double& vx, double& vy, double& omega,
   }
   else
   {
-    vx *= ratio_x;
+    vx *= ratio_x; 
     vy *= ratio_y;
     omega *= ratio_omega;
+
```
corot: ?

src/teb_local_planner_ros.cpp
```diff
@@ -869,13 +869,28 @@ double TebLocalPlannerROS::estimateLocalGoalOrientation(const std::vector<geomet
...
+      ratio_x = - min_vel_x / vx;
+  }
+  else // forward driving
+  {
+    if (vx > max_vel_x)
+      ratio_x = max_vel_x / vx;
+    else if (vx < min_vel_x)
+    {
+      ratio_x =   min_vel_x / vx;
+    }
```
corot: please adhere to current code style 
```suggestion
      ratio_x = min_vel_x / vx;
```

src/teb_local_planner_ros.cpp
```diff
@@ -869,13 +869,28 @@ double TebLocalPlannerROS::estimateLocalGoalOrientation(const std::vector<geomet
...
-  
+  double ratio_x = 1, ratio_omega = 1, ratio_y = 1, ratio_x_min = 1;
+
+  // backward driving
+  if(vx < 0)
+  {  
+    if (vx < -max_vel_x_backwards)
+      ratio_x = - max_vel_x_backwards / vx;
+    else if (vx > -min_vel_x)
+      ratio_x = - min_vel_x / vx;
```
corot:
```suggestion
      ratio_x = -max_vel_x_backwards / vx;
    else if (vx > -min_vel_x)
      ratio_x = -min_vel_x / vx;
```

corot (changes requested):

AlperenKosem:
requested changes are done

AlperenKosem:
@corot Could you review again?

cfg/TebLocalPlannerReconfigure.cfg
```diff
@@ -92,6 +92,10 @@ grp_robot.add("max_vel_x", double_t, 0,
 	"Maximum translational velocity of the robot",
 	0.4, 0.01, 100)   
 
+grp_robot.add("min_vel_x", double_t, 0, 
+	"Minimum translational velocity of the robot",
+	0.00, 0.00, 100)
```
corot, line 97:
```suggestion
	0.0, 0.0, 100)   
```

src/teb_local_planner_ros.cpp
```diff
@@ -869,12 +869,25 @@ double TebLocalPlannerROS::estimateLocalGoalOrientation(const std::vector<geomet
 }
       
       
-void TebLocalPlannerROS::saturateVelocity(double& vx, double& vy, double& omega, double max_vel_x, double max_vel_y, double max_vel_theta, double max_vel_x_backwards) const
+void TebLocalPlannerROS::saturateVelocity(double& vx, double& vy, double& omega, double max_vel_x, double max_vel_y, double max_vel_theta, double max_vel_x_backwards, double min_vel_x) const
```
corot, line 872: same order everywhere
```suggestion
void TebLocalPlannerROS::saturateVelocity(double& vx, double& vy, double& omega, double max_vel_x, double min_vel_x, double max_vel_y, double max_vel_theta, double max_vel_x_backwards) const
```

src/teb_local_planner_ros.cpp
```diff
@@ -869,12 +869,25 @@ double TebLocalPlannerROS::estimateLocalGoalOrientation(const std::vector<geomet
...
+      ratio_x = -max_vel_x_backwards / vx;
+    else if (vx > -min_vel_x)
+      ratio_x = -min_vel_x / vx;
+  }
+  else // forward driving
+  {
+    if (vx > max_vel_x)
+      ratio_x = max_vel_x / vx;
+    else if (vx < min_vel_x)
+      ratio_x = min_vel_x / vx;
```
corot, line 889: I'm not sure about this; ratio_x will get a value > 1, but that will be discarded [here](https://github.com/rst-tu-dortmund/teb_local_planner/pull/329/files#diff-a98cf96ec035dcabf1f5f3ba912efd8deaac734fd506f2b30f3f574f9512f3b3R910) if using proportional saturation (which is the preferred option, and so we will move below min_vel_x

corot (changes requested):
Works only when use_proportional_saturation is false (see my comment below)

src/teb_local_planner_ros.cpp
AlperenKosem: I didn't think proportional saturation side. Therefore i thought that there is no problem but you are right. If proportional saturation contains min velocity feature all twist components must speed up. Increased speed can causes unstability. Maybe this section can hard coded as below otherwise i have no solution for proportional saturation for min speed feature.
(I wrote pseudocode for explaining. When moved backwards, vx < min_vel_vx does not valid.)
```
if (cfg_.robot.use_proportional_saturation)
{
    double ratio = std::min(std::min(ratio_x, ratio_y), ratio_omega);
    vx *= ratio;
    if (vx < min_vel_x)
        vx = min_vel_x
    vy *= ratio;
    omega *= ratio;
}
```

corot: Yes, for sure;  but I wonder if that may trigger other problems... :thinking: 
@croesmann , @RainerKuemmerle , would u mind to take a look at this solution and confirm it doesn't break the intended behavior of proportional saturation?

corot: But this will fail with negative x velocities, right?

RainerKuemmerle: I think with a min velocity that requires other components to speed up for keeping the curvature is problematic, i.e. proportional scaling. There are in general cases where, for example, vy would violate max_vel_y. In particular on an omni directional drive for which it is perfectly fine to have (0, vy, 0) as a target command.
For a differential drive, the min vel param should affect the wheels if the motivation is that below some velocity command, the wheels are not moving. In this case, it might make more sense to reason on the min rotational velocity on the individual wheels directly to minimize the error in the curvature to follow by the robot.

corot: Good point; I also read the old @croesmann's answer to the same issue and he agrees that a minimum velocity doesn't play well with TEB architecture.
If nothing against, I'll close this PR by now. 
@AlperenKosem, as suggested, the min wheels velocity makes more sense at controller level.
But thanks a lot for the contribution; please don't feel discouraged to do more in the future!

AlperenKosem: Firstly i constrained velocity at controller level side but in this case, robot could not able to follow the trajectory because wheel velocities did not match the velocities which generated by planner. Therefore i added to min. vel. constraint to planner side. But i understand possible problems, thanks for reviews.

PR #332, opened 2021-11-25
Add prevent_look_ahead_poses_near_goal to fix oscillation

corot:
Never look ahead to the last few poses of the plan in order to prevent oscillating forward and backward over the goal when the xy_goal_tolerance is small. Specified by new parameter prevent_look_ahead_poses_near_goal whose default is zero.

Cherry-picked from #180 for melodic and without the additional reconfigurable parameters, to avoid cluttering the RQT tool.
We can add one-by-one whatever parameters we find useful to, as we did with `control_look_ahead_poses`.

Thanks to @howardcochran for the contribution

### DIFF
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -66,6 +66,10 @@ grp_trajectory.add("control_look_ahead_poses", int_t, 0,
   "Index of the pose used to extract the velocity command",
   1, 1, 100)     
 
+grp_trajectory.add("prevent_look_ahead_poses_near_goal",   int_t,   0,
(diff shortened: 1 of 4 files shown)


siferati (approved):

bxc237:
@corot Hi, could you explain more about this value? Let's say `control_look_ahead_poses` was set to 3, what is the reasonable value for `prevent_look_ahead_poses_near_goal`?

corot:
> @corot Hi, could you explain more about this value? Let's say `control_look_ahead_poses` was set to 3, what is the reasonable value for `prevent_look_ahead_poses_near_goal`?

I cherry-picked from PR #180, where the problem is explained in more detail
in your case, if u set `prevent_look_ahead_poses_near_goal` to 3, that means that lookahead pose will be the 3rd one except near the pose, that it will the 2nd, then 1st

PR #342, opened 2021-12-16
[ROS2] Missing dependency on tf2_eigen

Timple:
Used here: https://github.com/rst-tu-dortmund/teb_local_planner/blob/ros2-master/teb_local_planner/CMakeLists.txt#L34


corot (approved):

SteveMacenski (approved):
:+1:

PR #346, opened 2022-01-11
add max vel linear

renan028:
# Problem Description 
When we use holonomic robots, we should limit the maximum linear velocity (L-inf, L-1 or L-2 norm), not only `max_vel_x` and `max_vel_y`, because they are coupled.

Otherwise, the optimization for `vx` and `vy` results in `max_vel_x` and `max_vel_y` in open / collision-free paths. The robot will always move diagonally because the resultant (linear velocity) is greater. However, the maximum velocity is coupled and bounded to the actual maximum linear velocity of the robot. This new constraint should be added to the optimizer.

The video below shows the robot moving diagonally and it fixes the orientation at the end:

https://user-images.githubusercontent.com/6097267/148993227-5773c195-a3c6-4ad6-ab5a-503c104d751b.mp4

And one can see that the behaviour of the robot is not what we expect for a holonomic robot (changing  orientation a lot, and moving diagonally):

https://user-images.githubusercontent.com/6097267/148996957-d1a8698e-b858-4eb1-b1a5-319d458dc292.mp4

# Solution
Add `max_vel_linear` for holonomic robots and add the constraint to the optimization.

Long path not moving diagonally:

https://user-images.githubusercontent.com/6097267/148996627-7f598de7-23f8-4c17-818f-28df3a3e92f6.mp4

Moving around:

https://user-images.githubusercontent.com/6097267/148996037-e0d0d8df-6e35-4735-b835-095ed0cad98f.mp4

### DIFF
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -143,8 +143,8 @@ grp_robot_omni.add("max_vel_y", double_t, 0,
   "Maximum strafing velocity of the robot (should be zero for non-holonomic robots!)",
   0.0, 0.0, 100)
 
-grp_robot_omni.add("max_vel_linear", double_t, 0, 
-  "Maximum linear velocity of the robot. Ignore this parameter for non-holonomic robots (max_vel_linear == max_vel_x).",
+grp_robot_omni.add("max_vel_trans", double_t, 0, 
+  "Maximum linear velocity of the robot. Ignore this parameter for non-holonomic robots (max_vel_trans == max_vel_x).",
   0.4, 0.01, 100)
 
 grp_robot_omni.add("acc_lim_y", double_t, 0, 
@@ -334,8 +334,8 @@ grp_optimization.add("obstacle_cost_exponent", double_t, 0,
 # Optimization/Omni
 grp_optimization_omni = grp_optimization.add_group("OptOmni", type="hide")
 
-grp_optimization_omni.add("norm_vel_lin", int_t, 0,
-	"Norm constraint for translational velocity (1: L1, 2: L2). For example, enforcing hypot(vx, xy) <= v_max_lin for L2",
+grp_optimization_omni.add("norm_vel_trans", int_t, 0,
+	"Norm constraint for translational velocity (1: L1, 2: L2). For example, enforcing hypot(vx, xy) <= vel_max_trans for L2",
 	1, 1, 2)
   
   
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -141,6 +141,10 @@ grp_robot_omni = grp_robot.add_group("Omnidirectional", type="hide")
 
 grp_robot_omni.add("max_vel_y", double_t, 0, 
   "Maximum strafing velocity of the robot (should be zero for non-holonomic robots!)",
+  0.0, 0.0, 100)
+
+grp_robot_omni.add("max_vel_linear", double_t, 0, 
+  "Maximum linear velocity of the robot. Ignore this parameter for non-holonomic robots (max_vel_linear == max_vel_x).",
   0.0, 0.0, 100) 
 
 grp_robot_omni.add("acc_lim_y", double_t, 0, 
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -145,7 +145,7 @@ grp_robot_omni.add("max_vel_y", double_t, 0,
 
 grp_robot_omni.add("max_vel_linear", double_t, 0, 
   "Maximum linear velocity of the robot. Ignore this parameter for non-holonomic robots (max_vel_linear == max_vel_x).",
-  0.0, 0.0, 100) 
+  0.4, 0.01, 100)
 
 grp_robot_omni.add("acc_lim_y", double_t, 0, 
   "Maximum strafing acceleration of the robot",
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -330,13 +330,6 @@ grp_optimization.add("weight_adapt_factor", double_t, 0,
 grp_optimization.add("obstacle_cost_exponent", double_t, 0,
 	"Exponent for nonlinear obstacle cost (cost = linear_cost * obstacle_cost_exponent). Set to 1 to disable nonlinear cost (default)",
 	1, 0.01, 100)
-
-# Optimization/Omni
-grp_optimization_omni = grp_optimization.add_group("OptOmni", type="hide")
-
-grp_optimization_omni.add("norm_vel_trans", int_t, 0,
-	"Norm constraint for translational velocity (1: L1, 2: L2). For example, enforcing hypot(vx, xy) <= vel_max_trans for L2",
-	1, 1, 2)
   
   
 # Homotopy Class Planner
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -331,6 +331,12 @@ grp_optimization.add("obstacle_cost_exponent", double_t, 0,
 	"Exponent for nonlinear obstacle cost (cost = linear_cost * obstacle_cost_exponent). Set to 1 to disable nonlinear cost (default)",
 	1, 0.01, 100)
 
+# Optimization/Omni
+grp_optimization_omni = grp_optimization.add_group("OptOmni", type="hide")
+
+grp_optimization_omni.add("norm_vel_lin", int_t, 0,
+	"Norm constraint for translational velocity (1: L1, 2: L2). For example, enforcing hypot(vx, xy) <= v_max_lin for L2",
+	1, 1, 2)
   
   
 # Homotopy Class Planner
--- a/include/teb_local_planner/g2o_types/edge_velocity.h
+++ b/include/teb_local_planner/g2o_types/edge_velocity.h
@@ -253,16 +253,16 @@ public:
     double omega = g2o::normalize_theta(conf2->theta() - conf1->theta()) / deltaT->estimate();
 
 
-    double max_vel_trans_remaining;
-    if (cfg_->optim.norm_vel_trans == 1) {
-      max_vel_trans_remaining = std::max(0.0, cfg_->robot.max_vel_trans - std::abs(vx));  // L1 norm
-    }
-    else {
-      max_vel_trans_remaining = std::sqrt(std::max(0.0, cfg_->robot.max_vel_trans * cfg_->robot.max_vel_trans - vx * vx)); // L2 norm
-    }
-    double max_vel_y = std::min(max_vel_trans_remaining, cfg_->robot.max_vel_y);
+    double max_vel_trans_remaining_y;
+    double max_vel_trans_remaining_x;
+    max_vel_trans_remaining_y = std::sqrt(std::max(0.0, cfg_->robot.max_vel_trans * cfg_->robot.max_vel_trans - vx * vx)); 
+    max_vel_trans_remaining_x = std::sqrt(std::max(0.0, cfg_->robot.max_vel_trans * cfg_->robot.max_vel_trans - vy * vy)); 
 
-    _error[0] = penaltyBoundToInterval(vx, -cfg_->robot.max_vel_x_backwards, cfg_->robot.max_vel_x, cfg_->optim.penalty_epsilon);
+    double max_vel_y = std::min(max_vel_trans_remaining_y, cfg_->robot.max_vel_y);
+    double max_vel_x = std::min(max_vel_trans_remaining_x, cfg_->robot.max_vel_x);
+    double max_vel_x_backwards = std::min(max_vel_trans_remaining_x, cfg_->robot.max_vel_x_backwards);
+
+    _error[0] = penaltyBoundToInterval(vx, -max_vel_x_backwards, max_vel_x, cfg_->optim.penalty_epsilon);
     _error[1] = penaltyBoundToInterval(vy, max_vel_y, 0.0); // we do not apply the penalty epsilon here, since the velocity could be close to zero
     _error[2] = penaltyBoundToInterval(omega, cfg_->robot.max_vel_theta,cfg_->optim.penalty_epsilon);
 
--- a/include/teb_local_planner/g2o_types/edge_velocity.h
+++ b/include/teb_local_planner/g2o_types/edge_velocity.h
@@ -252,16 +252,16 @@ public:
     double vy = r_dy / deltaT->estimate();
     double omega = g2o::normalize_theta(conf2->theta() - conf1->theta()) / deltaT->estimate();
 
-    double max_vel_linear = std::max(std::max(cfg_->robot.max_vel_x, cfg_->robot.max_vel_y), cfg_->robot.max_vel_linear);
+    double max_vel_trans = std::max(std::max(cfg_->robot.max_vel_x, cfg_->robot.max_vel_y), cfg_->robot.max_vel_trans);
 
-    double max_linear_vel_remaining;
-    if (cfg_->optim.norm_vel_lin == 1) {
-      max_linear_vel_remaining = max_vel_linear - std::abs(vx);  // L1 norm
+    double max_vel_trans_remaining;
+    if (cfg_->optim.norm_vel_trans == 1) {
+      max_vel_trans_remaining = max_vel_trans - std::abs(vx);  // L1 norm
     }
     else {
-      max_linear_vel_remaining = std::sqrt(max_vel_linear * max_vel_linear - vx * vx); // L2 norm
+      max_vel_trans_remaining = std::sqrt(max_vel_trans * max_vel_trans - vx * vx); // L2 norm
     }
-    double max_vel_y = std::min(max_linear_vel_remaining, cfg_->robot.max_vel_y);
+    double max_vel_y = std::min(max_vel_trans_remaining, cfg_->robot.max_vel_y);
 
     _error[0] = penaltyBoundToInterval(vx, -cfg_->robot.max_vel_x_backwards, cfg_->robot.max_vel_x, cfg_->optim.penalty_epsilon);
     _error[1] = penaltyBoundToInterval(vy, max_vel_y, 0.0); // we do not apply the penalty epsilon here, since the velocity could be close to zero
--- a/include/teb_local_planner/g2o_types/edge_velocity.h
+++ b/include/teb_local_planner/g2o_types/edge_velocity.h
@@ -251,9 +251,14 @@ public:
     double vx = r_dx / deltaT->estimate();
     double vy = r_dy / deltaT->estimate();
     double omega = g2o::normalize_theta(conf2->theta() - conf1->theta()) / deltaT->estimate();
-    
+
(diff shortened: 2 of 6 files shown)


cfg/TebLocalPlannerReconfigure.cfg
```diff
@@ -141,6 +141,10 @@ grp_robot_omni = grp_robot.add_group("Omnidirectional", type="hide")
 
 grp_robot_omni.add("max_vel_y", double_t, 0, 
   "Maximum strafing velocity of the robot (should be zero for non-holonomic robots!)",
+  0.0, 0.0, 100)
+
+grp_robot_omni.add("max_vel_linear", double_t, 0, 
+  "Maximum linear velocity of the robot. Ignore this parameter for non-holonomic robots (max_vel_linear == max_vel_x).",
   0.0, 0.0, 100)
```
siferati:
```suggestion
  0.4, 0.01, 100) 
```

S: Use the same values as `max_vel_x`.

include/teb_local_planner/g2o_types/edge_velocity.h
```diff
@@ -251,9 +251,14 @@ class EdgeVelocityHolonomic : public BaseTebMultiEdge<3, double>
     double vx = r_dx / deltaT->estimate();
     double vy = r_dy / deltaT->estimate();
     double omega = g2o::normalize_theta(conf2->theta() - conf1->theta()) / deltaT->estimate();
-    
+
+    double max_vel_linear = std::max(std::max(cfg_->robot.max_vel_x, cfg_->robot.max_vel_y), cfg_->robot.max_vel_linear);
+    // double max_linear_vel_left = std::sqrt(max_vel_linear * max_vel_linear - vx * vx); // L2
+    double max_linear_vel_left = max_vel_linear - std::abs(vx); // L1
```
siferati: Q: Why are we using L1 norm by default instead of L2?
S: Add a parameter to decide between using L1 or L2 norm.
R: Rename `max_linear_vel_left` to something else (e.g. `max_linear_vel_remaining`), as _left_ indicates direction and is a bit confusing!

include/teb_local_planner/teb_config.h
```diff
@@ -279,6 +280,7 @@ class TebConfig
     robot.cmd_angle_instead_rotvel = false;
     robot.is_footprint_dynamic = false;
     robot.use_proportional_saturation = false;
+    robot.max_vel_linear = robot.max_vel_x;
```
siferati: F: I don't think this does anything as it will be overwritten by the default value from dynamic reconfigure. So it's better to change this line to the same value (I suggested `0.4` above) to avoid confusion

cfg/TebLocalPlannerReconfigure.cfg
```diff
@@ -141,6 +141,10 @@ grp_robot_omni = grp_robot.add_group("Omnidirectional", type="hide")
 
 grp_robot_omni.add("max_vel_y", double_t, 0, 
   "Maximum strafing velocity of the robot (should be zero for non-holonomic robots!)",
+  0.0, 0.0, 100)
+
+grp_robot_omni.add("max_vel_linear", double_t, 0,
```
siferati:
```suggestion
grp_robot_omni.add("max_vel_trans", double_t, 0, 
```
S: rename from _linear_ to _translational_.

src/teb_config.cpp
```diff
@@ -350,6 +352,13 @@ void TebConfig::checkParameters() const
   // weights
   if (optim.weight_optimaltime <= 0)
       ROS_WARN("TebLocalPlannerROS() Param Warning: parameter weight_optimaltime shoud be > 0 (even if weight_shortest_path is in use)");
+
+  // holonomic check
+  if (robot.max_vel_y >= 0) {
```
siferati:
```suggestion
  if (robot.max_vel_y > 0) {
```
F: it's 0 by default, so this check always happens.

src/teb_config.cpp
```diff
@@ -350,6 +352,13 @@ void TebConfig::checkParameters() const
...
+    if (robot.max_vel_linear < robot.max_vel_x || robot.max_vel_linear < robot.max_vel_y) {
+      ROS_WARN("TebLocalPlannerROS() Param Warning: max_vel_linear < max_vel_x or max_vel_linear < max_vel_y. Assuming max(max_vel_x, max_vel_y)");
```
siferati: S: instead of assuming `max(max_vel_x, max_vel_y)` I think it would be better to stick with the given value and simply print the warning. Vx and Vy would be clamped in `saturateVelocity` to match the given max linear vel, instead of the other way around.

src/teb_local_planner_ros.cpp
```diff
@@ -903,6 +905,15 @@ void TebLocalPlannerROS::saturateVelocity(double& vx, double& vy, double& omega,
     vy *= ratio_y;
     omega *= ratio_omega;
   }
+
+  max_vel_linear = std::max(std::max(max_vel_x, max_vel_y), max_vel_linear);
+  double vel_linear = std::sqrt(vx*vx + vy*vy);
```
siferati:
```suggestion
  double vel_linear = std::hypot(vx, vy);
```
S: use `std:hypot` instead

src/teb_local_planner_ros.cpp
```diff
@@ -903,6 +905,15 @@ void TebLocalPlannerROS::saturateVelocity(double& vx, double& vy, double& omega,
...
+  if (vel_linear > std::abs(max_vel_linear))
```
siferati:
```suggestion
  if (vel_linear > max_vel_linear)
```
F: no need for `abs` since this willl always be >= 0

siferati (changes requested):

cfg/TebLocalPlannerReconfigure.cfg
renan028: linear velocity is vastly more common in the literature. Is there a specific reason you want to change it to translational?

include/teb_local_planner/g2o_types/edge_velocity.h
renan028: It is way cheaper, and the behaviour is the one expected.
Sometimes optimization cannot keep up with L2 and the behaviour is not the one expected. I will add the parameter though.

corot:
I think this can be interesting also for Nav2 version, I please, @SteveMacenski , take a look if u have time

cfg/TebLocalPlannerReconfigure.cfg
siferati: Translational seems to be the common name to use in ROS planners (e.g. [dwa](http://wiki.ros.org/dwa_local_planner) and [eband](http://wiki.ros.org/eband_local_planner))

corot: that's the name on DWA,,, and also we have below `trans_stopped_vel`

include/teb_local_planner/teb_local_planner_ros.h
```diff
@@ -354,9 +354,11 @@ class TebLocalPlannerROS : public nav_core::BaseLocalPlanner, public mbf_costmap
    * @param max_vel_y Maximum strafing velocity (for holonomic robots)
    * @param max_vel_theta Maximum (absolute) angular velocity
    * @param max_vel_x_backwards Maximum translational velocity for backwards driving
+   * @param max_vel_linear Maximum translational velocity holonomic robots
    */
   void saturateVelocity(double& vx, double& vy, double& omega, double max_vel_x, double max_vel_y,
-                        double max_vel_theta, double max_vel_x_backwards) const;
+                        double max_vel_theta, double max_vel_x_backwards, 
+                        double max_vel_linear) const;
```
corot:
```suggestion
                        double max_vel_linear, double max_vel_theta, double max_vel_x_backwards) const;
```

src/teb_config.cpp
```diff
@@ -84,6 +84,7 @@ void TebConfig::loadRosParamFromNodeHandle(const ros::NodeHandle& nh)
   nh.param("is_footprint_dynamic", robot.is_footprint_dynamic, robot.is_footprint_dynamic);
   nh.param("use_proportional_saturation", robot.use_proportional_saturation, robot.use_proportional_saturation);
   nh.param("transform_tolerance", robot.transform_tolerance, robot.transform_tolerance);
+  nh.param("max_vel_linear", robot.max_vel_linear, robot.max_vel_linear);
```
corot: Put together with the other vel params, and always in a fixed order (e.g. after max_vel_y and before max_vel_theta)

src/teb_config.cpp
```diff
@@ -213,6 +214,7 @@ void TebConfig::reconfigure(TebLocalPlannerReconfigureConfig& cfg)
   robot.wheelbase = cfg.wheelbase;
   robot.cmd_angle_instead_rotvel = cfg.cmd_angle_instead_rotvel;
   robot.use_proportional_saturation = cfg.use_proportional_saturation;
+  robot.max_vel_linear = cfg.max_vel_linear;
```
corot: same

src/teb_local_planner_ros.cpp
```diff
@@ -421,7 +421,8 @@ uint32_t TebLocalPlannerROS::computeVelocityCommands(const geometry_msgs::PoseSt
   
   // Saturate velocity, if the optimization results violates the constraints (could be possible due to soft constraints).
   saturateVelocity(cmd_vel.twist.linear.x, cmd_vel.twist.linear.y, cmd_vel.twist.angular.z,
-                   cfg_.robot.max_vel_x, cfg_.robot.max_vel_y, cfg_.robot.max_vel_theta, cfg_.robot.max_vel_x_backwards);
+                   cfg_.robot.max_vel_x, cfg_.robot.max_vel_y, cfg_.robot.max_vel_theta, cfg_.robot.max_vel_x_backwards, 
+                   cfg_.robot.max_vel_linear );
```
corot: after max_vel_y

corot (changes requested):
still requires some refinements, but nice fix!

@SteveMacenski, I think this can be interesting also for Nav2 version, so please, take a look if u have time

cfg/TebLocalPlannerReconfigure.cfg
```diff
@@ -334,8 +334,8 @@ grp_optimization.add("obstacle_cost_exponent", double_t, 0,
 # Optimization/Omni
 grp_optimization_omni = grp_optimization.add_group("OptOmni", type="hide")
 
-grp_optimization_omni.add("norm_vel_lin", int_t, 0,
-	"Norm constraint for translational velocity (1: L1, 2: L2). For example, enforcing hypot(vx, xy) <= v_max_lin for L2",
+grp_optimization_omni.add("norm_vel_trans", int_t, 0,
```
corot: Q: what's the point of this?  what's the difference between enforcing L1 or L2?
To me u always want to enforce that the magnitude of vector vx, vy is <= max_vel_trans

corot (changes requested):
lgtm, but I would remove the L1, L2 param

include/teb_local_planner/g2o_types/edge_velocity.h
```diff
@@ -251,9 +251,20 @@ class EdgeVelocityHolonomic : public BaseTebMultiEdge<3, double>
     double vx = r_dx / deltaT->estimate();
     double vy = r_dy / deltaT->estimate();
     double omega = g2o::normalize_theta(conf2->theta() - conf1->theta()) / deltaT->estimate();
-    
+
+    double max_vel_trans = std::max(std::max(cfg_->robot.max_vel_x, cfg_->robot.max_vel_y), cfg_->robot.max_vel_trans);
```
siferati: F: remove this and use `cfg_->robot.max_vel_trans` as per https://github.com/rst-tu-dortmund/teb_local_planner/pull/346#discussion_r782843978 (will need to clamp `max_vel_trans_remaining` to 0.0 to account for the cases when `vx` > `max_vel_trans`)

siferati (changes requested):

cfg/TebLocalPlannerReconfigure.cfg
siferati: See https://github.com/rst-tu-dortmund/teb_local_planner/pull/346#discussion_r782823162

> It [L1] is way cheaper, and the behaviour is the one expected.
Sometimes optimization cannot keep up with L2 and the behaviour is not the one expected.

include/teb_local_planner/g2o_types/edge_velocity.h
```diff
@@ -251,9 +251,20 @@ class EdgeVelocityHolonomic : public BaseTebMultiEdge<3, double>
...
+    }
+    else {
+      max_vel_trans_remaining = std::sqrt(max_vel_trans * max_vel_trans - vx * vx); // L2 norm
+    }
+    double max_vel_y = std::min(max_vel_trans_remaining, cfg_->robot.max_vel_y);
+
     _error[0] = penaltyBoundToInterval(vx, -cfg_->robot.max_vel_x_backwards, cfg_->robot.max_vel_x, cfg_->optim.penalty_epsilon);
-    _error[1] = penaltyBoundToInterval(vy, cfg_->robot.max_vel_y, 0.0); // we do not apply the penalty epsilon here, since the velocity could be close to zero
+    _error[1] = penaltyBoundToInterval(vy, max_vel_y, 0.0); // we do not apply the penalty epsilon here, since the velocity could be close to zero
     _error[2] = penaltyBoundToInterval(omega, cfg_->robot.max_vel_theta,cfg_->optim.penalty_epsilon);
```
corot, line 267: S / Q:  cannot just add a constraint here like hypot(vx, vy) <= max_vel_trans?  I think it should work, and will be way simpler  :thinking:

corot (changes requested):
Q

include/teb_local_planner/g2o_types/edge_velocity.h
renan028: Do you mean to add a new `_error[3]`?
For that, we need to change the optimization, by changing the matrix information size [here](https://github.com/rst-tu-dortmund/teb_local_planner/blob/melodic-devel/src/optimal_planner.cpp#L757), the `EdgeVelocityHolonomic` template to accommodate a new variable,  and by adding a new variable to optimize (`vel_trans`, which is the coupled velocities). I have tried that, but the behavior is very similar to the current L2 implementation.

corot: but,,,  isn't much cleaner than the added implementation?  what is weird is that your constraint seems to subordinate vy to vx,,, what is not the intended behavior
adding _error[3] would be impartial

renan028: hum, yes. I think the optimizer will try to favour Vx and subordinate Vy on each step.
However, I am just thinking about the "new" variable, and we cannot add that because it is not controllable. `vel_trans` is a consequence of Vx and Vy and we cannot optimize that variable.
But, I can fix the subordinate thing with the following implementation:
```cpp
    double max_vel_trans_remaining_y;
    double max_vel_trans_remaining_x;
    if (cfg_->optim.norm_vel_trans == 1) { // L1 Norm
      max_vel_trans_remaining_y = std::max(0.0, cfg_->robot.max_vel_trans - std::abs(vx));  
      max_vel_trans_remaining_x = std::max(0.0, cfg_->robot.max_vel_trans - std::abs(vy));  
    }
    else { // L2 Norm
      max_vel_trans_remaining_y = std::sqrt(std::max(0.0, cfg_->robot.max_vel_trans * cfg_->robot.max_vel_trans - vx * vx)); 
      max_vel_trans_remaining_x = std::sqrt(std::max(0.0, cfg_->robot.max_vel_trans * cfg_->robot.max_vel_trans - vy * vy)); 
    }

    double max_vel_y = std::min(max_vel_trans_remaining_y, cfg_->robot.max_vel_y);
    double max_vel_x = std::min(max_vel_trans_remaining_x, cfg_->robot.max_vel_x);
    double max_vel_x_backwards = std::min(max_vel_trans_remaining_x, cfg_->robot.max_vel_x_backwards);

    _error[0] = penaltyBoundToInterval(vx, -max_vel_x_backwards, max_vel_x, cfg_->optim.penalty_epsilon);
    _error[1] = penaltyBoundToInterval(vy, max_vel_y, 0.0); // we do not apply the penalty epsilon here, since the velocity could be close to zero
    _error[2] = penaltyBoundToInterval(omega, cfg_->robot.max_vel_theta,cfg_->optim.penalty_epsilon);
```
I have just tested and the behavior is nice for both L1 and L2
What do you think?

corot: mmm...  better;  
btw, is L1 and L2 behavior anyhow different?  cause as I see this, there's only one correct behavior (what I mention before with the additional constraint)

renan028: the behavior is the same for both L1 and L2.
I will remove L1 since even the computation is the same now and L2 will always try to optimize to the actual max_vel_trans (L2)

src/teb_config.cpp
```diff
@@ -350,6 +352,13 @@ void TebConfig::checkParameters() const
   // weights
   if (optim.weight_optimaltime <= 0)
       ROS_WARN("TebLocalPlannerROS() Param Warning: parameter weight_optimaltime shoud be > 0 (even if weight_shortest_path is in use)");
+
+  // holonomic check
+  if (robot.max_vel_y > 0) {
+    if (robot.max_vel_trans < robot.max_vel_x || robot.max_vel_trans < robot.max_vel_y) {
+      ROS_WARN("TebLocalPlannerROS() Param Warning: max_vel_trans < max_vel_x or max_vel_trans < max_vel_y.");
```
corot: S: maybe we can add more details about why is this bad

src/teb_config.cpp
```diff
@@ -350,6 +352,13 @@ void TebConfig::checkParameters() const
...
+    }
+  }
```
corot, line 365: can also add another warning if trans is bigger than both  -->  explain that the robot will tend to rotate to maximize both velocity components

corot (approved):
cool!

siferati (approved):

src/teb_config.cpp
```diff
@@ -350,6 +352,17 @@ void TebConfig::checkParameters() const
   // weights
   if (optim.weight_optimaltime <= 0)
       ROS_WARN("TebLocalPlannerROS() Param Warning: parameter weight_optimaltime shoud be > 0 (even if weight_shortest_path is in use)");
+
+  // holonomic check
+  if (robot.max_vel_y > 0) {
+    if (robot.max_vel_trans < robot.max_vel_x || robot.max_vel_trans < robot.max_vel_y) {
```
corot: Here we can use min, right?

renan028: yes

src/teb_config.cpp
```diff
@@ -350,6 +352,17 @@ void TebConfig::checkParameters() const
   // weights
   if (optim.weight_optimaltime <= 0)
       ROS_WARN("TebLocalPlannerROS() Param Warning: parameter weight_optimaltime shoud be > 0 (even if weight_shortest_path is in use)");
+
+  // holonomic check
+  if (robot.max_vel_y > 0) {
+    if (robot.max_vel_trans < std::min(robot.max_vel_x, robot.max_vel_trans)) {
+      ROS_WARN("TebLocalPlannerROS() Param Warning: max_vel_trans < max_vel_x or max_vel_trans < max_vel_y. Note that vel_trans = sqrt(Vx^2 + Vy^2), thus max_vel_trans will limit Vx and Vy in the optimization step.");
```
corot:
```suggestion
      ROS_WARN("TebLocalPlannerROS() Param Warning: max_vel_trans < min(max_vel_x, max_vel_y). Note that vel_trans = sqrt(Vx^2 + Vy^2), thus max_vel_trans will limit Vx and Vy in the optimization step.");
```

corot:
There was an old issue for this #89

Issue #348, opened 2022-01-18
base_max_vel_x and the other base_* parameters are not declared


tuxnese:
When using the speed limit filter I have noticed that the max speed of the robot was 0.4, no matter the value of max_vel_x. 
Looking at the code of the setSpeedLimit, I see:

cfg_->robot.max_vel_x = cfg_->robot.base_max_vel_x;

But the base_max_vel_x parameter and the other base_* parameters are not declared, so it is not possible to change their values. Please, correct me if I am wrong.

Why not to use directly the parameters without the base prefix?
Otherwise the base* parameters should be declared to allow their modification.

tuxnese:
I think I got why the base* parameters are not declared. The robot base variables are just used to store the value of the max_vel_x and the other corresponding variables and they are set each time the parameters are set.
The problem is that this does not work when setting the values of the parameters at the loading of the node. So the base* variables keep their initial values. The workaround is to set the parameters once the nodes are up.

tuxnese:
PR to fix it : https://github.com/rst-tu-dortmund/teb_local_planner/pull/349

PR #349, opened 2022-01-18
[ROS2] Fix max speed when using speed limit filter

tuxnese:
This PR is meant to fix issue  [#348](https://github.com/rst-tu-dortmund/teb_local_planner/issues/348) .
I have tested, it works.
Please do not mind the first two commits.


tuxnese:
Are there news about this PR? I still see same problem on the last version of TEB

PR #350, opened 2022-01-18
[ROS2] Fix unit tests

guilyx:
- Uncommented both tests in CMake.
- Rearrange the HomotopyClass test to check if the goal has been reached instead of running it forever.


guilyx:
@SteveMacenski not sure if this is useful since there's no CI set up for Rolling here, I can put up the tests results log

PR #355, opened 2022-02-11
[fix] only limit with max_vel_trans for holonomic vehicles

remco-r:
a new parameter `max_vel_trans` was added in commit [f737130](https://github.com/rst-tu-dortmund/teb_local_planner/commit/f737130fa6a9024fef9fcdca6eadacaa69bf30a4) 

According to the documentation it should not affect for non-holonomic robots. But this was not implemented, and my robots vel_x was limited nevertheless.
`Maximum linear velocity of the robot. Ignore this parameter for non-holonomic robots (max_vel_trans == max_vel_x).`

### DIFF
--- a/src/teb_config.cpp
+++ b/src/teb_config.cpp
@@ -214,6 +214,11 @@ void TebConfig::reconfigure(TebLocalPlannerReconfigureConfig& cfg)
   robot.wheelbase = cfg.wheelbase;
   robot.cmd_angle_instead_rotvel = cfg.cmd_angle_instead_rotvel;
   robot.use_proportional_saturation = cfg.use_proportional_saturation;
+  if (cfg.max_vel_trans == 0.0)
+  {
+    ROS_INFO_STREAM("max_vel_trans is not set, setting it equal to max_vel_x: " << robot.max_vel_x);
+    cfg.max_vel_trans = robot.max_vel_x;
+  }
   robot.max_vel_trans = cfg.max_vel_trans;
   
   // GoalTolerance
--- a/cfg/TebLocalPlannerReconfigure.cfg
+++ b/cfg/TebLocalPlannerReconfigure.cfg
@@ -89,7 +89,7 @@ grp_viapoints.add("via_points_ordered",   bool_t,   0,
 grp_robot = gen.add_group("Robot", type="tab")
 
 grp_robot.add("max_vel_x", double_t, 0, 
-	"Maximum translational velocity of the robot",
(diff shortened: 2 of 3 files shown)


corot:
mmm...  the idea was to have a top limit for all linear velocities, overruling both x and y max vel!
but the comment is completely wrong. I propose to replace this PR with rewriting the comment with

"Maximum linear velocity of the robot, overruling both max_vel_x, max_vel_y and its linear combination."

renan028:
> mmm... the idea was to have a top limit for all linear velocities, overruling both x and y max vel! but the comment is completely wrong. I propose to replace this PR with rewriting the comment with
> 
> "Maximum linear velocity of the robot, overruling both max_vel_x, max_vel_y and its linear combination."

I see. The problem is that, for non-holonomic robots, we now need to set both `max_vel_x` and `max_vel_trans` to the same value, and as @corot said the comment is wrong.
I suggest to change the default value of  `max_vel_trans` to `0` or a negative value, and check the value inside the parameter reconfiguration. If  `max_vel_trans <= 0` do `max_vel_trans = max_vel_x`.
@remco-r what do you think?

remco-r:
We agree the parameter documentation has to be updated. I included a commit to do that.

I can imagine that many users forget to set the `max_vel_trans` variable, and try to figure out why their robot is going slow (as i did). Especially people with a running configuration that update the package. And with this fix it does become backwards compatible for the non-holonomic case. So i have included a commit for this as well.

src/teb_config.cpp
```diff
@@ -74,7 +74,7 @@ void TebConfig::loadRosParamFromNodeHandle(const ros::NodeHandle& nh)
   nh.param("max_vel_x", robot.max_vel_x, robot.max_vel_x);
   nh.param("max_vel_x_backwards", robot.max_vel_x_backwards, robot.max_vel_x_backwards);
   nh.param("max_vel_y", robot.max_vel_y, robot.max_vel_y);
-  nh.param("max_vel_trans", robot.max_vel_trans, robot.max_vel_trans);
+  nh.param("max_vel_trans", robot.max_vel_trans, robot.max_vel_trans); // todo
```
corot: todo?

corot (approved):
lgtm

src/teb_config.cpp
remco-r: removed

siferati (approved):

renan028 (approved):

PR #362, opened 2022-04-10
fix misc header

AlbertKisa:


corot:
Sorry, reformatting an entire code base is not sensible PR
- Authors chose the stile they wanted
- This renders git history nearly useless

PR #363, opened 2022-04-20
Gitignore JetBrains projects

corot:
Something else we should git-ignore?


siferati:
Maybe just exclude all hidden files (except for stuff like gitignore ofc)? Otherwise we'll have to list multiple files per IDE (e.g. `.vscode` for vscode stuff)

```
# Hidden files for different IDE (exclude this file)
!.gitignore
.*
```

siferati (approved):

PR #371, opened 2022-07-10
Fix build failure in ros2 humble with nav2 1.1.0

chenjunnn:


corot (approved):
Working with latest nav2 sources

I suppose this will break compilation with pre-humble distros
Maybe we should create a Galactic branch from current master before merging?
@SteveMacenski, thoughts?

SteveMacenski (approved):

SteveMacenski:
Done and done!

PR #383, opened 2022-09-22
[Bugfix] fixed init teb not clear poses vec

2linkthefire:
1. fixed init when teb time diff vec is empty, but pose vec may has pose causing core dump.

### DIFF
--- a/src/timed_elastic_band.cpp
+++ b/src/timed_elastic_band.cpp
@@ -326,6 +326,7 @@ bool TimedElasticBand::initTrajectoryToGoal(const PoseSE2& start, const PoseSE2&
 {
   if (!isInit())
   {   
+    clearTimedElasticBand();
     addPose(start); // add starting point
     setPoseVertexFixed(0,true); // StartConf is a fixed constraint during optimization
 
@@ -391,6 +392,7 @@ bool TimedElasticBand::initTrajectoryToGoal(const std::vector<geometry_msgs::Pos
   
   if (!isInit())
   {
+    clearTimedElasticBand();
     PoseSE2 start(plan.front().pose);
     PoseSE2 goal(plan.back().pose);
     


2linkthefire:
@corot

corot:
> 1. fixed init when teb time diff vec is empty, but pose vec may has pose causing core dump.

Hi, thanks for the fix. I never experience that bug. How can I reproduce it?

2linkthefire:
> > 1. fixed init when teb time diff vec is empty, but pose vec may has pose causing core dump.
> 
> Hi, thanks for the fix. I never experience that bug. How can I reproduce it?

Okay, in general, teb will first add a pose when initializing, and then add a pose and a time_diff in pairs, so the number of time_diff will be one less than a pose; 
Because teb provides interfaces similar to deletePose() and deleteTimeDiff(), there may be only one pose left, which will be judged as False by TimedElasticBand::isInit(), and the last pose is not cleared when reinitializing, and this dirty data will cause the difference between the number of poses and the time_diff greater than 1; 
When using pose and time_diff in pairs, it may crash;

corot:
> Okay, in general, teb will first add a pose when initializing, and then add a pose and a time_diff in pairs, so the number of time_diff will be one less than a pose; Because teb provides interfaces similar to deletePose() and deleteTimeDiff(), there may be only one pose left, which will be judged as False by TimedElasticBand::isInit(), and the last pose is not cleared when reinitializing, and this dirty data will cause the difference between the number of poses and the time_diff greater than 1; When using pose and time_diff in pairs, it may crash;

Sounds reasonable. But still I don't know how to reproduce. Even better would be a unit test that reproduces the problem. Would it be possible to add one together with the fix?  If not, clear instructions to reproduce it would do.

2linkthefire:
> Sounds reasonable. But still I don't know how to reproduce. Even better would be a unit test that reproduces the problem. Would it be possible to add one together with the fix? If not, clear instructions to reproduce it would do.

Thanks, I took a closer look at the comments on the min_samples parameter and the related deletions from native code, the number of poses should not be removed to less than min_samples number, so the problem does not exist in origin code.

corot:
> Thanks, I took a closer look at the comments on the min_samples parameter and the related deletions from native code, the number of poses should not be removed to less than min_samples number, so the problem does not exist in origin code.

Nice!  can we close this PR, then?   And thanks for the investigation!

PR #385, opened 2022-10-05
More accurate visualization of obstacles

siferati:
Obstacle markers had a fixed size of 10cm, but this means that if the costmap has a low resolution and the robot is navigating close to obstacles, in rviz it will seem like the robot is in collision even though it isn't.

This PR changes it so that the obstacles have the same size as the costmap resolution.
This should provide a much more accurate visualization of the actual obstacle size!


corot (approved):
Minor comment: I would rename `scale` param as `size`

PR #393, opened 2022-12-16
fix segfault at robot-model update (#255 revisited)

corot:
Resurrecting #255:

We found exactly the same issue #245, reproduced in the same way, and I can confirm that @dorezyuk's PR #255 fixes it

I open a new PR because:
- Need to solve some conflicts
- I also want to patch noetic and ros2
- I'm not sure about the need for commit f4c43757fc59f647ae7f5c305cb7036056764934

About @amakarow's [comment](https://github.com/rst-tu-dortmund/teb_local_planner/pull/255#issuecomment-754089623), I think the code becomes cleaner now (if nothing else, we have almost 4 to 1 deletions!)

@dorezyuk, can u explain why the lock is needed?  Thanks, and thanks for the fix!

### DIFF
--- a/include/teb_local_planner/g2o_types/edge_dynamic_obstacle.h
+++ b/include/teb_local_planner/g2o_types/edge_dynamic_obstacle.h
@@ -80,7 +80,7 @@ public:
   }
   
   /**
-   * @brief Construct edge and specify the time for its associated pose (neccessary for computeError).
+   * @brief Construct edge and specify the time for its associated pose (necessary for computeError).
    * @param t_ Estimated time until current pose is reached
    */      
   EdgeDynamicObstacle(double t) : t_(t)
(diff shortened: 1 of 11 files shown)


siferati:
We need the lock because otherwise you're modifying a field of the config object outside the protection of the mutex.
The other fields are protected by the mutex inside `TebConfig::reconfigure`, but `robot_model` is not since you set it outside in `TebLocalPlannerROS::reconfigureCB`.

It's the same problem we had with DWA a while ago

corot:
> We need the lock because otherwise you're modifying a field of the config object outside the protection of the mutex. The other fields are protected by the mutex inside `TebConfig::reconfigure`, but `robot_model` is not since you set it outside in `TebLocalPlannerROS::reconfigureCB`.
> 
> It's the same problem we had with DWA a while ago

added

siferati (approved):

zst1406217:
This is the exact problem I have in my research. When I reconfigure the parameters, the segfault occurs occasionally, but not from the beginning, this has confused me a long time. I prove this is helpful!

corot:
> This is the exact problem I have in my research. When I reconfigure the parameters, the segfault occurs occasionally, but not from the beginning, this has confused me a long time. I prove this is helpful!

Thanks for reporting. Merged

PR #394, opened 2022-12-23
Show also the actual footprint at infeasible poses

corot:
Useful to detect when the footprint model is not accurate enough
I chose not to use a parameter, as either visualization can be disabled by clicking the marker namespace, as shown in the video:
- red: footprint model (as before)
- orange: actual footprint

https://user-images.githubusercontent.com/322610/217402422-613ab552-db5b-4005-9b7c-f699f93ef6d4.mp4

### DIFF
--- a/src/visualization.cpp
+++ b/src/visualization.cpp
@@ -119,8 +119,8 @@ void TebVisualization::publishLocalPlanAndPoses(const TimedElasticBand& teb) con
 
 
 
-void TebVisualization::publishRobotFootprintModel(const PoseSE2& current_pose, const BaseRobotFootprintModel& robot_model, const std::string& ns,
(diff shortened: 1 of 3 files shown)


src/visualization.cpp
```diff
@@ -129,8 +129,8 @@ void TebVisualization::publishRobotFootprintModel(const PoseSE2& current_pose, c
   robot_model.visualizeRobot(current_pose, markers, color);
   if (markers.empty())
     return;
-  
-  int idx = 1000000;  // avoid overshadowing by obstacles
+
+  int idx = 0;
```
corot, line 133: OOS: not needed, as it uses its own namespace

siferati (approved):

PR #401, opened 2023-02-08
Apply penalty_epsilon to max_vel_y on EdgeVelocityHolonomic

corot:
This is more a question than a contribution. Why exactly we don't apply penalty_epsilon to max_vel_y?
The comment doesn't make much sense to me, as this is an holonomic edge and so we should not prioritize any orientation

Thanks in advance for any input!


siferati (approved):

corot:
Closing in favor of #404

PR #414, opened 2023-08-16
Fix exceptions for Nav2 Iron

tonynajjar:
TEB doesn't compile with Nav2 Iron and more recent (because of https://github.com/ros-planning/navigation2/pull/3227). This PR fixes this


tonynajjar:
Merging this will probably break a working setup for many that are on Humble, should a new iron branch be created? @doisyg @SteveMacenski @corot

Rayman:
Thank you for the PR! It compiles on my system. I'll test this some more this week.

1. (optional) A humble branch could be created from the current branch
2. This PR should be merged into master

corot:
Thanks @tonynajjar and sorry for the late handling
I merged @Rayman's matching PR, so this is not needed. I'm ignoring Iron anyway, as this month is EOL

PR #415, opened 2023-10-30
fixed obstacle select in obstacle edge

2linkthefire:
1. fixed select closest obstacle from left and right side.

### DIFF
--- a/src/optimal_planner.cpp
+++ b/src/optimal_planner.cpp
@@ -510,7 +510,7 @@ void TebOptimalPlanner::AddEdgesObstacles(double weight_multiplier)
             continue;
           
           // determine side (left or right) and assign obstacle if closer than the previous one
-          if (cross2d(pose_orient, obst->getCentroid()) > 0) // left
+          if (cross2d(pose_orient, obst->getCentroid() - teb_.Pose(i).position()) > 0) // left
           {
               if (dist < left_min_dist)
               {


2linkthefire:
> @corot
> Can you help review/merge this PR?

We need more details, like:
- what's the problem you are addressing
- current behavior
- expected behavior
- ideally a unit test

2linkthefire:
In the ` AddEdgesObstacles() `, there is a logic to filter the two nearest obstacles on the left and right;
```
          // determine side (left or right) and assign obstacle if closer than the previous one
          if (cross2d(pose_orient, obst->getCentroid()) > 0) // left
```

When judging left and right, the unit vector of the pose orientation and **the relative position of the obstacle and pose** should be used for cross2d calculation, but it is obvious that the coordinates of obs in the local TF are currently used `obst->getCentroid()`.


I use Marker to visualize all the obstacles on the left side. It can be seen that in the second quadrant of the coordinate system, when the pose is in the same X-axis direction, all obstacles are considered to be on the left side. When the pose is in the opposite direction, all obstacles are considered as on the right side. ;
image
image


This BUG may cause obstacles to be ignored, affecting the centering of the pose.

2linkthefire:
> > @corot
> > Can you help review/merge this PR?
> 
> We need more details, like:
> 
> * what's the problem you are addressing
> * current behavior
> * expected behavior
> * ideally a unit test

Thanks for the reply, I just posted the detailed reason.

corot:
@2linkthefire, sorry for the delay. Let me see if I got the problem right:
On this situation, the robot points to the north, and so all the obstacles should be on the left

image

But w/o your fix they are considered as on the right. So it actually fixes the issue.

Am I correct?

corot:
@SteveMacenski , the issue [is also present on ROS 2](https://github.com/rst-tu-dortmund/teb_local_planner/blob/ros2-master/teb_local_planner/src/optimal_planner.cpp#L515).

corot:
@2linkthefire, I double checked with another person, and we both agree that the fix is valid. Merging

